PR #230, opened 2014-01-23
Add JSONP support

timjacobi:
I worked with a different JSON-API today which made me realise that Eve doesn't have support for JSONP yet. With this feature you're able to add a `JSONP_ARGUMENT` to your config. When you request a resource you can add this argument and the response will be wrapped in a JavaScript function call. The obviously only works with JSON. If XML is requested the setting is ignored.

Example

request: `curl -i http://localhost:5000/?callback=JSONP_CALLBACK`
response: `JSONP_CALLBACK( # json response here )`

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -393,6 +393,11 @@ uppercase.
 
 ``DOMAIN``                      A dict holding the API domain definition.
                                 See `Domain Configuration`_.
+
+``JSONP_ARGUMENT``              This option will cause the response to be
+                                wrapped in a JavaScript function call if the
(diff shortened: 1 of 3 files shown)


papaonlegs:
++ to jsonp

nicolaiarocci:
Hey folks, thanks for hanging in there for so long. This has finally been merged and is scheduled for 0.6 release. See https://github.com/nicolaiarocci/eve/commit/1442953f523a7141339d07416274f866fa0ce753.

Cheers!

timjacobi:
Great, thank you!

PR #489, opened 2014-10-22
Added serialization support for nested fields using dot notation.

opticode:
This change is to support PATCH commands containing updates to nested fields in subdocuments, using dot notation as in MongoDB. An example would be a PATCH like this:

```{"data.fan": "5435c2a4d302e926e818b2d5"}```

The idea is to traverse the schemas of each field in the dotted name to find the appropriate field_schema to use for serialization.

### DIFF
--- a/eve/methods/common.py
+++ b/eve/methods/common.py
@@ -312,8 +312,8 @@ def serialize(document, resource=None, schema=None, fields=None):
             elif '.' in field:
                 # handle nested "dot notation" field names (e.g. "a.b.c")
                 # todo: this code currently requires a schema for each part in
-                #       the dotted name, i.e. b must be in a's schema & c in b's
-                #       schema
+                #       the dotted name, i.e. b must be in a's schema & c in
+                #       b's schema
                 parts = field.split('.')
                 p_schema = schema
                 try:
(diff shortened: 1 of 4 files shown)


nicolaiarocci:
I took  a different approach: dotted documents are normalised before validation is performed, so validation and updates can be performed seamlessly. Also, this will allow for dotted notation to be supported on all write/edit methods: POST, PUT and PATCH. See https://github.com/nicolaiarocci/eve/commit/75060b4ff476a918bed0b00073802e6a19452262.

Thanks!

opticode:
Awesome Nicola, I was just discussing this possibility with a colleague a few days ago. Thanks for not forgetting about this issue.

opticode:
Regarding the replacement of the entire sub-document, which is especially bothersome for PATCHes: could a possible solution be to first retrieve the sub-document, then modify it, and then continue with the patch with this "full" sub-document?

nicolaiarocci:
Right now both normal and dotted notations have the same update patterns:

    {"location": {"city": "a nested city"}}

Will replace the subdocument as the the dotted notation would. I felt like adding the warning because somehow the dotted notation seems misleading to me.

Atomic subdocument field updates could probably be added for dotted notation, thus diverging from normal notation behaviour, but that would require some extra work at either the business layer (merge the original subdocument with the changes) or at the data layer level (use atomic update operators). The former is probably preferable, as it would allow non-mongo data layers to take advantage of the new feature. Food for thought :smile:

For the time being I kind of like to have the two notations behave the same way for the time being. Let's see how it is

PR #532, opened 2014-12-17
Add support for multiple origins when using CORS

joshvillbrandt:
This pull request adds support for multiple origins when using CORS headers.

Motivation:

I have multiple web sites (multiple domains) that utilize one common Eve API. If I access a resource through the first site, everything is fine, but if I access that same resource through the second site, I gets a CORS HTTP error because the browser has cached the response sent from the first site (`Access-Control-Allow-Origin: site1`.)

The [W3C CORS recommendation page](http://www.w3.org/TR/cors/#resource-implementation) suggests that hosts intending to serve to multiple origins should either return `Access-Control-Allow-Origin: *` or, if the origin is specified in that header, to add a `Vary: Origin` header. I think the easiest option is option #1 and that is what is implemented in this pull request.

### DIFF
--- a/eve/render.py
+++ b/eve/render.py
@@ -195,7 +195,8 @@ def _prepare_response(resource, dct, last_modified=None, etag=None,
         methods = app.make_default_options_response().headers.get('allow', '')
 
         if '*' in domains:
-            resp.headers.add('Access-Control-Allow-Origin', '*')
+            resp.headers.add('Access-Control-Allow-Origin', origin)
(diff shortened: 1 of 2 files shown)


joshvillbrandt:
Hey all,

This solution still doesn't work for me when using Google Chrome. I'm investigating setting the `Access-Control-Allow-Credentials` header per this [Chrome ticket](https://code.google.com/p/chromium/issues/detail?id=409090), but I'm not sure that is really what I want.

Has anyone else run into this?

PR #546, opened 2015-01-15
Update features.rst

msabramo:


nicolaiarocci:
Thanks!

nicolaiarocci:
Hm not sure why it is not flagged as merged by @GitHub.. but it actually is :smile:

PR #548, opened 2015-01-17
Fix bug when embedded is used on subdocument with missing field

touilleMan:
I've run into a bug when a get request is done with a embedded params on a missing subdocument

###### Example
Considering a request `GET http://api.lvh.me:8080/elements?projection={"tests.test": 1}`
If the `elements` collection contains an element which doesn't have a `tests` entry, the request is going to crash leading to a 500 response.


nicolaiarocci:
thanks!

PR #552, opened 2015-01-27
Fix ObjectId value in ID field of type string.

girogiro:
The config.ID_FIELD field doesn't respect the *query_objectid_as_string* setting and always cast the values in lookup to ObjectId type. If an item's ID field contains a string that is a valid ObjectId value, then lookup by that value searches by ObjectId and doesn't find the item.

This PR fixes the bug and handles the config.ID_FIELD in accordance with *query_objectid_as_string* setting relying on eventual cast to ObjectId in _mongotize() method.

I am sorry for not adding corresponding tests. It needs a new test configuration with ID_FIELD of type string and was too difficult for me.


nicolaiarocci:
This has been merged, thank you! Please try not to submit PRs from your `develop` branch but create a dedicate one (or then github has trouble auto-flagging the PR as merged)

PR #553, opened 2015-01-30
Fix mongo Validator to work without transparent_schema_rules=True

touilleMan:
Hi,

I've found initializing mongo's Validator with `transparent_schema_rules=True` is  quite error-prone
For exemple considering a schema with a typo (`child_field` not enclosed in `{'type': 'dict', 'schema': {...}}`):
```
{
    'field': {
        'type': 'list',
        'schema': {
            'child_field': {'type': 'string'}
        }
    }
}
```
With `transparent_schema_rules=True` this schema is considered as valid and any arbitrary payload will be silently accepted !

This fix remove the need to use `transparent_schema_rules=True`... and fix a unittest whose invalid schema were silently failing, ouch ! ;-)

### DIFF
--- a/eve/io/mongo/validation.py
+++ b/eve/io/mongo/validation.py
@@ -49,7 +49,7 @@ class Validator(Validator):
         self.resource = resource
         self._id = None
         self._original_document = None
-        super(Validator, self).__init__(schema, transparent_schema_rules=True)
+        super(Validator, self).__init__(schema)
         if resource:
             self.allow_unknown = config.DOMAIN[resource]['allow_unknown']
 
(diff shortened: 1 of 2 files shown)


nicolaiarocci:
Thanks!

kynan:
How about adding a configuration option to allow this (either globally or per resource)? We have a rather unusual use case: "dynamic schemas" i.e. we read the schema for some resources from an API endpoint. A schema obviously doesn't validate a schema, so without `transparent_schema_rules` we need to use a custom validator without countless "fake" noop validation rules.

kynan:
@nicolaiarocci I've implemented a way to enable `transparent_schema_rules` for the `eve.io.mongo.Validator` on a global or per-resource basis, see #740.

PR #557, opened 2015-02-13
Add support for serving media files from GridFS and return file urls in media fields

aikoven:
This PR introduces `RETURN_MEDIA_AS_URL` settings entry, an alternative to `RETURN_MEDIA_AS_BASE64_STRING`. It enables returning media fields as URLs to endpoint that serves files from MediaStorage.

It currently works with GridFS, but I need some help to find the most convenient way to make it agnostic of MediaStorage type.


nicolaiarocci:
First of all this is cool stuff, thank you.

On making it agnostic of MediaStorage type: my first thought is to add a `media_endpoint` (or just `endpoint`) method to `eve.io.media.MediaStorage` base class. Then `GridFSStorage` would simply provide the implementation like it does with `get` and other methods. With that done you would then need to refactor the endpoints module to use the new `app.media.endpoint` method.

Adding support for a non-CRUD method to a Storage class is not super elegant though. An alternative would be adding a `MediaEndpoint` base class to the `eve.io.media` module, but that wold require a new argument to be passed around at startup, which does not make a lot of sense since you're likely going to be using the same media/storage technology.

Thoughts?

PS: would be nice to hear @amleczko opinion on this one

aikoven:
I think that we can go without adding a method to `MediaStorage` class, because if we can obtain a file descriptor then the remaining code for endpoint is the same for all types. So the problem here is to convert endpoint argument to something we can pass to `MediaStorage#get` method.

Also I noticed that `Mongo(DataLayer)` performs this conversion using `self._mongotize` in its interface methods; maybe it would be convenient to place analogous conversion into `GridFSMediaStorage` methods. Then `media_endpoint` would be generic for all kinds of storages.

amleczko:
That's a very nice feature! 
Having a separate base class for `MediaEndpoint` seems over-killing. I would keep it as simple as possible. I like the idea of adding `endpoint` to `eve.io.media.MediaStorage` class - it seems more elegant.
Is there a reason why `MEDIA_ENDPOINT` is none by default?

nicolaiarocci:
This has been merged in https://github.com/nicolaiarocci/eve/commit/b38d014c4f077642b85b0471cc4debc0c01c2f71 with some additional work (besides documentation) to make sure that endpoint code remains agnostic of storage engine. Also ``MEDIA_ENDPOINT`` now defaults to `media` so by simply setting ``RETURN_MEDIA_AS_URL`` you switch the magic on. 

Please review it and PR/discuss any further improvement.

nicolaiarocci:
PS: Thanks!

PR #558, opened 2015-02-14
Typo in exemple "Endpoint-level Authentication"?

kumy:
Is this a typo in the example ?
Doing as showed, give me:

```
node = {
    'authentication': NodeAuth,
[...]
}


TypeError: unbound method get_request_auth_value() must be called with NodeAuth instance as first argument (got nothing instead)
```

Adding parenthesis, fix the problem:
```
node = {
    'authentication': NodeAuth(),
[...]
}

### DIFF
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -102,7 +102,7 @@ resource-level ``authentication`` setting when we are defining the API
 
     DOMAIN = {
         'people': {
-            'authentication': MySuperCoolAuth,
+            'authentication': MySuperCoolAuth(),
             ...
             },
         'invoices': ... 


nicolaiarocci:
On which Eve version are you? I think we had this issue in 0.3 or 0.4 but can't check right now.

kumy:
I used 0.5 before, and ugraded 0.5.1 before reporting.

nicolaiarocci:
Ah great it was actually a subtle bug that you spotted there. Should be fixed now, thank you so much!

kumy:
glad to be useful

PR #559, opened 2015-02-15
Add `X-Total-Count` header for collection GET response

aikoven:
Added custom header containing total count of items in response payload. This is handy for HEAD requests when user wants to know items count without retrieving response body.

A particular use case is to get the count of unread posts using `where` query without loading posts themselves.


nicolaiarocci:
This has been merged in https://github.com/nicolaiarocci/eve/commit/07b02569a67ead9d0b595edd0779f4da8fb96345 - thanks!

PR #561, opened 2015-02-16
Fix for database concurrency issues.

kapouille:
Ensure we handle database concurrency in case there are concurrent accesses to the database that modify the same objects.
The current implementation will ensure "_etag" is consistent and retry the ``PUT``/``PATCH`` operation if not.

The diff is messy as there's been a lot of indented lines, but in a nutshell the ``PUT``/``PATCH`` operations are surrounded by a ``while True`` loop that's ``continue``d if there's a mid-air collision, otherwise the loop carries on in execution and either ``throw``s or ``return``s.

There was a suggested use of ``find_and_modify()``, but I couldn't find a solution using this function beside the use of etag that would allow us **not** to modify the database if there was indeed a collision.  I also noted when I tried this solution that there was somehow date mismatches between what ``find_and_modify()`` returns and the original document which would cause comparisons to fail (some would be offset-aware while other wouldn't).

Let me know what you think.

### DIFF
--- a/eve/io/base.py
+++ b/eve/io/base.py
@@ -78,6 +78,9 @@ class DataLayer(object):
        the _datasource helper function has been added.
     """
 
+    class OriginalChangedError(Exception):
+        pass
+
     # if custom serialize functions are needed, add them to the 'serializers'
     # dictionary, eg:
     # serializers = {'objectid': ObjectId, 'datetime': serialize_date}
@@ -198,7 +201,7 @@ class DataLayer(object):
         """
         raise NotImplementedError
 
-    def update(self, resource, id_, updates):
+    def update(self, resource, id_, updates, original):
         """ Updates a collection/table document/row.
         :param resource: resource being accessed. You should then use
                          the ``_datasource`` helper function to retrieve
@@ -206,17 +209,24 @@ class DataLayer(object):
         :param id_: the unique id of the document.
         :param updates: json updates to be performed on the database document
                         (or row).
+        :param original: definition of the json document that should be
+        updated.
+        :raise OriginalChangedError: raised if the database layer notices a
+        change from the supplied `original` parameter.
         """
         raise NotImplementedError
 
-    def replace(self, resource, id_, document):
+    def replace(self, resource, id_, document, original):
         """ Replaces a collection/table document/row.
         :param resource: resource being accessed. You should then use
                          the ``_datasource`` helper function to retrieve
                          the actual datasource name.
         :param id_: the unique id of the document.
         :param document: the new json document
-
+        :param original: definition of the json document that should be
+        updated.
+        :raise OriginalChangedError: raised if the database layer notices a
+        change from the supplied `original` parameter.
         .. versionadded:: 0.1.0
         """
         raise NotImplementedError
--- a/eve/io/mongo/mongo.py
+++ b/eve/io/mongo/mongo.py
@@ -375,7 +375,7 @@ class Mongo(DataLayer):
                 filter_, changes, **self._wc(resource))
 
             if result and result["n"] == 0:
-                raise self.OriginalChangedError
+                raise self.OriginalChangedError()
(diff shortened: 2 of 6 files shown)


kapouille:
Updated the implementation to work even when etags aren't enabled. When performing an update, the query uses the original to match. If the original isn't matched, the whole operation is retried.

rs:
Am I the only one not to understand what this PR is trying to solve? :)

The `while True` scares me a lot though.

kapouille:
Imagine your Eve app is hosted on more than one thread. 

Imagine that 2 web clients attempt to PUT/PATCH the same object at precisely the same instant and that those requests are handled by two different threads. 

Then there's a considerable chances that one of the 2 requests will be overwritten by the other one without any notification of the collision back to the user. This is particularly critical when using etags as the its guarantees are not enforced in that particular instance.

kapouille:
With regards to the ``while True``, a counter could be implemented, but it really only loops when there's a database collision. I'm happy to use another flow control mechanism.

kapouille:
By the way, this is more than hypothetical. This regularly happened to our application before we put such a fix in.

rs:
In my understanding, if such collision happen with etag activated, stored etag would have changed for the second thread, and as a consequence, the `If-Match` precondition should fail. Instead of a `while True`, I guess you should return a 412 error and let the API client to handle it.

kapouille:
Yes, but we're talking about the code in the method handler. This is not atomic, and in between the time ``get_document`` and the time ``app.data.update`` or ``app.data.replace`` is called, another thread could have updated the object you're meant to handle. In such case, nothing will be caught, and no 412 will be raised.

I invite you to consult ``eve.methods.patch_internal`` and ``eve.methods.put_internal`` where you will be able to see that gap.

rs:
I understand the issue, but what prevents you from raising the 412 instead of retrying?

kapouille:
Right! Now I understand your remark. Of course, it would make sense to delegate handling this to the client. I shall update my code to do that instead.

pfreixes:
This is a good try to minimize or reduce to zero the length of code path that can get document inconsistency because of the concurrency. As I can see pymongo [1] handles access from various threads therefore this commit should fix any issue about inconsistency.

IMHO @rs is right, the client should be warned about the inconsistency and delegate the decision to him.

As I can see this patch uses a filter using all fields to detect this inconsistency, the filter should be better with the `_etag` field for performance reasons.

[1] http://api.mongodb.org/python/current/faq.html#is-pymongo-thread-safe

kapouille:
@pfreixes If you look at 39e1b13 and before, you'll notice that I used to only do a filter on ``_etag``. I realise that in fact enforcing consistency really only makes sense in the context of having etags enabled, so I'll probably revert that as well.

kapouille:
The filter is now set on matching ``_etag`` when relevant and 412 raised as soon as there's a conflict (which means no more ``while True``)

nicolaiarocci:
Thanks for this. Could  you please rebase on top of `develop` and ping when it's done?

PS: I like how this code evolved since the initial submission. Thanks @rs and @pfreixes for your precious feedback.

kapouille:
@nicolaiarocci : this is now rebased.

nicolaiarocci:
This has been merged in https://github.com/nicolaiarocci/eve/commit/992e7a71f800579dc06220e7e6b3f2436385b616. Had to rebase once more since I had pushed a few docs updates in the meantime :)

Thank you!

kiniou:
Hello,

I'm a bit late to comment on this PR, but the addition of 'original' argument in Datalayer API methods replace() and update() may be worth a note in the changelog of 0.5.2 since it breaks extensions like eve-elastic or eve-sqlalchemy.

nicolaiarocci:
:+1:

nicolaiarocci:
Note added. Thanks.

PR #563, opened 2015-02-22
For nicola/read only validation on replace

andersson:
This series of patches makes it possible to PUT an object with read-only properties, given that these keep their original values.


andersson:
Sorry, seems I didn't re-run all the test cases after the default-changes; I will have another look, either drop the default parts (and handle the lack of default values in the added test cases) or make the tests pass again.

nicolaiarocci:
Closing for the time being.

PR #564, opened 2015-02-23
Make it possible to PUT objects with read-only parts

andersson:
This series introduce the possibility of PUTing an object with read only fields, given that those fields pass the readonly check.


nicolaiarocci:
Thanks! It's merged in https://github.com/nicolaiarocci/eve/commit/39fba5a1c5be7b95b629b9dab91fee742a0446dd

(had to rebase it on develop)

Issue #524, opened 2014-12-08
VERSIONING with item_lookup_field causes KeyError


otibsa:
When using `VERSIONING=True` and `'item_lookup_field': 'lastname'`, I cannot GET a specific version of my endpoint:

`GET .../people/obama` works, but `...people/obama?version=1` creates the following error:

```
Traceback (most recent call last):
  File "/foobar/eve0.5_env/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1836, in __call__
    return self.wsgi_app(environ, start_response)
  File "/foobar/eve0.5_env/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1820, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/foobar/eve0.5_env/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1403, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/foobar/eve0.5_env/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/foobar/eve0.5_env/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/foobar/eve0.5_env/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/foobar/eve0.5_env/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/foobar/eve0.5_env/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/foobar/eve0.5_env/lib/python2.7/site-packages/Eve-0.5_dev-py2.7.egg/eve/endpoints.py", line 90, in item_endpoint
    response = getitem(resource, **lookup)
  File "/foobar/eve0.5_env/lib/python2.7/site-packages/Eve-0.5_dev-py2.7.egg/eve/methods/common.py", line 237, in rate_limited
[...]

PR #582, opened 2015-03-26
Fix item_lookup_field for versioned resources

otibsa:
See #524.

I don't know how the tests for this need to be implemented.
Would be great, if I could get some help there.

What I tested manually:
```python
...
VERSIONING = True
ITEM_URL = 'regex("[a-z0-9]{0,24}")'
DOMAIN = {
    'people' = {
        'item_lookup_field': 'lastname',
        'schema': {
            'firstname': {'type': 'string'},
            'lastname': {'type': 'string', 'required': True, 'unique': True}
        }
    }
}
```

```
GET /people/obama
GET /people/obama?version=1
GET /people/obama?version=diffs

### DIFF
--- a/eve/methods/get.py
+++ b/eve/methods/get.py
@@ -238,8 +238,13 @@ def getitem(resource, **lookup):
 
     if version == 'all' or version == 'diffs':
         # find all versions
-        lookup[versioned_id_field()] = lookup[app.config['ID_FIELD']]
-        del lookup[app.config['ID_FIELD']]
+        item_lookup_field = \
+            app.config['DOMAIN'][resource].get('item_lookup_field')
+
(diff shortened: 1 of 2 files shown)


nicolaiarocci:
@joshvillbrandt would be great if you could lend your hand here: versioning is your baby :smile:

joshvillbrandt:
Ref #524.

nicolaiarocci:
So what is the state of this PR at the time being? I think tests are still missing and we're still waiting on @joshvillbrandt code review, right?

otibsa:
That was my understanding as well. As I said, I did some "manual" tests, but I don't know how to implement them in the testing suite.

PR #587, opened 2015-04-01
Create failing test for replace subdocument bug

amagdas:


nicolaiarocci:
Could you base this on top of `master` instead of develop? So we can more easily start a 0.5.4 branch in which this code will be included. Thanks!

amagdas:
Replaced by #589

Issue #588, opened 2015-04-01
On PATCH, subdocuments are always merged


amagdas:
Currently there is no way to remove a key from a subdocument after it has been added, not even using the dot notation.
Added a failing test for this: https://github.com/nicolaiarocci/eve/pull/587
This got introduced with this commit: https://github.com/nicolaiarocci/eve/commit/026ce09c4ee1f595edc401f94778d39c5bce3742
I  don't think this is the expected behavior for PATCH requests...

amagdas:
@nicolaiarocci Maybe we should revert https://github.com/nicolaiarocci/eve/commit/026ce09c4ee1f595edc401f94778d39c5bce3742?

amagdas:
Actually, I think the proper solution is to keep only readonly fields from the subdocument when resolving the payload to be used by PATCH, I will update my PR with this fix.

amagdas:
Provided a fix for this in #587 , please let me know if it's ok.

nicolaiarocci:
hmm on second thought, how do you exactly remove a field form the main document with PATCH, I don't think it is possible... so why should we change the behaviour for subdocuments? https://github.com/nicolaiarocci/eve/commit/026ce09c4ee1f595edc401f94778d39c5bce3742 was introduced exactly because subdocs were being replaced, while the desired behaviour was to be able to replace subdocuments fields instead.

amagdas:
https://github.com/nicolaiarocci/eve/commit/026ce09c4ee1f595edc401f94778d39c5bce3742 was added so read-only fields in subdocuments won't get replaced, at least that's what I understood from the discussion.
I don't think that's a good behavior to not be able to remove fields from the subdocuments if the field is not read-only, it's quite useful to do so, especially for storing data having a tabular like format.

nicolaiarocci:
That commit also made sure that nested document fields were not being overwritten if not explictly mentioned, as you can see in this test: https://github.com/nicolaiarocci/eve/blob/develop/eve/tests/methods/patch.py#L403. 

While your proposed change does not overwrite readonly fields, it still allows for other fields to be deleted if not explicitly included with the payload, as we can see in your new test (https://github.com/amagdas/eve/blob/replace-subdoc/eve/tests/methods/patch.py#L214). Of course you also had to amend the previously mentioned test, setting the `int`  field to readonly to prevent deletion.

I understand from where you are coming (tabular data), but how about using PUT for these scenarios? I also wonder what @pfreixes would think about this change.

[...]

PR #589, opened 2015-04-01
fix(patch) - keep read-only fields when patching subdocuments

amagdas:
Replaces #587


amagdas:
The heisenbug is still there...

nicolaiarocci:
See discussion in https://github.com/nicolaiarocci/eve/issues/588

Issue #591, opened 2015-04-03
Geopoint coordinates does not accept integers


Tethik:
The coordinates have to be floats for eve to accept them as valid input. This is a problem with e.g. javascript, which seems to automatically convert e.g. 1.0 to 1 when sending via json. Other floats do not seem to have this problem.

I'll see if I can fix this and return with a pull request.

PoC
===

app.py
```
from eve import Eve

app = Eve()
app.run()
```

settings.py
```
schema = {
    # Schema definition, based on Cerberus grammar. Check the Cerberus project
    # (https://github.com/nicolaiarocci/cerberus) for details.
    'firstname': {
        'type': 'string',
        'minlength': 1,
        'maxlength': 10,
    },

    'weight': {
        'type': 'float',
    },

    'location': {
        'type': 'point'
    },
}


people = {
    # 'title' tag used in item links. Defaults to the resource title minus
    # the final, plural 's' (works fine in most cases but not for 'people')
    'item_title': 'person',

    # most global settings can be overridden at resource level
    'resource_methods': ['GET', 'POST'],

    'schema': schema
}

DOMAIN = {
    'people': people,
}
```

```
curl -d '[{"location": { "type": "Point", "coordinates": [1, 1.0]}, "firstname": "barack", "weight": 1}]' -H 'Content-Type: application/json'  http://127.0.0.1:5000/people
```
Results in:
```
{"_error": {"code": 422, "message": "Insertion failure: 1 document(s) contain(s) error(s)"}, "_status": "ERR", "_issues": {"location": "Point not correct {'coordinates': [1, 1.0], 'type': 'Point'}: "}}
```

Whereas
```
curl -d '[{"location": { "type": "Point", "coordinates": [1.1, 1.0]}, "firstname": "barack", "weight": 1}]' -H 'Content-Type: application/json'  http://127.0.0.1:5000/people
```
returns OK.

Tethik:
While I was at it I also noticed this weird code:
```python
factory = getattr(sys.modules[__name__], geometry["type"])
factory(geometry)
```
Passing user input to **getattr** then executing it also feels like it could lead to some nasty security hole down the line. 

Currently abusing this behaviour lets you insert arbitary data into a GeometryCollection document's geometries-field bypassing validation. E.g. 
```
[...]

PR #592, opened 2015-04-03
Geopoint decimal fix

Tethik:
See #591 

Fixed by adding new test and allowing for integers in the validation typecheck.

One minor caveat is that integers will also be stored as integers and as such a different type in mongo.
However, querying still works fine. See screenshot below.

2015-04-03-235836_681x513_scrot

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -47,6 +47,7 @@ Patches and Contributions
 - Javier Gonel
 - Jean Boussier
 - Jen Montes
+- Joakim Uddholm
 - Joseph Heck
 - Johan Bloemberg
 - John Deng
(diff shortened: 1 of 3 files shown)


nicolaiarocci:
:+1:

nicolaiarocci:
Thanks!

Tethik:
Np. Note there is still minor improvements which can be made to the validation by looking at the geojson specification. Currently eve does not differentiate between polygon and multilinestring validation. Polygon is supposed to have lines starting and ending on the same points, whereas multilinestring carries no such requirements.

Ref: http://geojson.org/geojson-spec.html#geometry-objects

Maybe I'll fix this later.

PR #601, opened 2015-04-14
Overwrite embedded_fields settings by 0 in ?embedded-query

otibsa:
```python
'foobar': {
    embedded_fields = ["field1"],
    ...
}
```
A `GET .../foobar?embedded={"field1": 0}` now does not embed `field1`, so the query-setting is prioritized over the DOMAIN-setting.

This way we can have embedded fields by default and turn off the embedding at will by using the embedded-query.

### DIFF
--- a/eve/tests/methods/get.py
+++ b/eve/tests/methods/get.py
@@ -632,6 +632,22 @@ class TestGet(TestBase):
         content = json.loads(r.get_data())
         self.assertFalse('missing-field' in content['_items'][0])
 
+        # Test default fields to be embedded
+        invoices['embedded_fields'] = ['person']
+        r = self.test_client.get("%s/" % invoices['url'])
+        self.assert200(r.status_code)
+        content = json.loads(r.get_data())
+        self.assertTrue('location' in content['_items'][0]['person'])
+
+        # Test that default fields are overwritten by ?embedded=...0
+        invoices['embedded_fields'] = ['person']
+        embedded = '{"person": 0}'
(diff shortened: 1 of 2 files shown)


eve/tests/methods/get.py
```diff
@@ -632,6 +632,22 @@ def test_get_embedded(self):
...
 
+        # Test default fields to be embedded
+        invoices['embedded_fields'] = ['person']
+        r = self.test_client.get("%s/" % invoices['url'])
+        self.assert200(r.status_code)
+        content = json.loads(r.get_data())
+        self.assertTrue('location' in content['_items'][0]['person'])
+
+        # Test that default fields are overwritten by ?embedded=...0
+        invoices['embedded_fields'] = ['person']
```
nicolaiarocci: this is probably redundant

otibsa: You're right, further down (new line 748) also

nicolaiarocci:
Overall this is a cool addition. I wonder however if there are legit reasons for the maintainer *not* wanting clients to be abale to exclude auto-embedded fields. Thoughts?

otibsa:
At least in my use case, there is no reason for "pushing" fields to the client. Besides, a client can always choose to drop fields of a response.

nicolaiarocci:
Thanks this has been merged in https://github.com/nicolaiarocci/eve/commit/0a9d473924dfca0ed576a81b4dcbca91889c2c56

PR #603, opened 2015-04-16
Feature timezone

nypgand1:
Store datetime as GMT,  parse and render it as ``DATE_TIMEZONE`` in configuration. Use [pytz] (http://pytz.sourceforge.net/) to implement the feature, and it allow values in [List of tz database time zones] (http://en.wikipedia.org/wiki/List_of_tz_database_time_zones).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -67,6 +67,7 @@ Patches and Contributions
 - Kracekumar
 - Kurt Bonne
 - Kurt Doherty
+- Lin, Wei-Hao
 - Magdas Adrian
 - Marc Abramowitz
 - Marcus Cobden
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -154,9 +154,15 @@ uppercase.
                                     stored as ``datetime`` values. In
                                     responses, ``datetime`` values will be
                                     rendered as JSON strings using this format.
-                                    Defaults to the RFC1123 (ex RFC 822)
-                                    standard ``a, %d %b %Y %H:%M:%S GMT``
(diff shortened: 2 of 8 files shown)


nicolaiarocci:
This is cool, and I am willing to give it a spin. However, it currently breaks the test suite.

nypgand1:
@nicolaiarocci It seems like it is the ``%Z`` part of ``DATE_FORMAT`` that makes it break. 

In the [doc of time.strptime] (https://docs.python.org/2/library/time.html#time.strptime), it says:
> Support for the %Z directive is based on the values contained in tzname and whether daylight is true. Because of this, it is platform-specific except for recognizing UTC and GMT which are always known (and are considered to be non-daylight savings timezones).

I replaces the default value of ``DATE_FORMAT`` by``%a, %d %b %Y %H:%M:%S %Z`` because the original one is ``%a, %d %b %Y %H:%M:%S GMT``.

Any suggestions?

ddurieux:
It make sense to store date in GMT in DB, but not sure it's the role of backend (EVE) to convert to a timezone. For me, it's the role of frontend to convert to user timezone

nicolaiarocci:
Yeah, overall I think it is better remain timezone-naive (also the issue is not easily solved without some hacking http://stackoverflow.com/questions/526406/python-time-to-age-part-2-timezones). We should consider that most REST APIs are probably hit by clients in all timezones anyway. @nypgand1 thanks!

nypgand1:
Thank you guys! At least learn something on this issue :)

Issue #354, opened 2014-06-01
POST on sub resources doesn't set the "route" parameters


rs:
Sub resources with GET method append the `url` parameters to the lookup but on POST or PUT method, those parameters aren't set.

For the following sub request:

```
invoices = {
    'url': 'people/<regex("[a-f0-9]{24}"):contact_id>/invoices'
    ...
```

A POST on `/people/1234…/invoices` should set the `contact_id` field to `1234…` so created invoice is automatically associated with the parent resource.

PR #609, opened 2015-04-20
Apply Route Params to Documents Before Validation

seglberg:
Route parameters are applied to new documents before they are validated.
This ensures that documents with required fields will be populated
before they are validated.

See nicolaiarocci/eve#354

---

Added a new test case to test the required field scenario.
All tests passed with both Python 2.7 and 3.4.

---

Pings:
@dexteradeus


nicolaiarocci:
Thanks, this has been merged in https://github.com/nicolaiarocci/eve/commit/4b567f3963d65e29bbd298f74a5a4b56c3202099

PR #612, opened 2015-04-22
make sure original is not modified during patch

### LINKED ISSUE #611: original in on_updated callback is having nested document updates
hey,

atm `original` in `on_update` can have different content than `original` in `on_updated`

it should be enough to use `deepcopy` in [patch](https://github.com/nicolaiarocci/eve/blob/develop/eve/methods/patch.py#L173)

petrjasek:
so that `on_update` and `on_updated` callbacks will get
same values

fix #611

### DIFF
--- a/eve/methods/patch.py
+++ b/eve/methods/patch.py
@@ -10,6 +10,7 @@
     :license: BSD, see LICENSE for more details.
 """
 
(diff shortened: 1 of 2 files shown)


nicolaiarocci:
Thanks!

PR #613, opened 2015-04-22
Invalidate cache when _latest_version changes in versioned doc.

### LINKED ISSUE #597: Cache is not invalidated when _latest_version changes
@nicolaiarocci,

So I guess I lied when I said there were no impacts to document versioning in #369. Since some fields are synthesized on document retrieval (notably the `_latest_version` attribute), comparing the if-match etag against the etag stored in the document will yield incorrect 304 responses.

Possible fixes:

1. update stored etags for old versions when a new version is saved
1. give up mongo's native projections in favor of performing the projection in the data store abstraction layer
1. make two data store queries for every document - only to get the etag and one to get the projection

Option 1 seems like the path of least resistance. Thoughts?

I still have some investigation to do to prove that this is a problem in the first place, but I wanted to document my progress somewhere because it is dinner time! (I'll either find an existing test or create a new test and will update this ticket with the results.)

nckpark:
Closes #597.

The last modified and etag data from the latest document version
should be used when determining if a cached document version is
still valid. Otherwise, 304 Not Modified responses will be sent
when the _latest_version field has changed.

### DIFF
--- a/eve/methods/get.py
+++ b/eve/methods/get.py
@@ -240,10 +240,10 @@ def getitem(resource, **lookup):
         cache_valid = (last_modified <= req.if_modified_since)
         cache_validators[cache_valid] += 1
     if req.if_none_match:
-        # If-None-Match etag check important since Last-Modified dates have
-        # less than 1 second resolution
-        cache_valid = (etag == req.if_none_match)
-        cache_validators[cache_valid] += 1
+        if (resource_def['versioning'] is False) or \
+           (document['_version'] == document['_latest_version']):
+            cache_valid = (etag == req.if_none_match)
+            cache_validators[cache_valid] += 1
     # If all cache validators are true, return 304
     if (cache_validators[True] > 0) and (cache_validators[False] == 0):
         return {}, last_modified, etag, 304
--- a/eve/methods/get.py
+++ b/eve/methods/get.py
@@ -209,11 +209,9 @@ def getitem(resource, **lookup):
     latest_doc = None
     cursor = None
 
-    # calculate last_modified and retrieve latest etag before get_old_document
-    # rolls back the document, allowing us to invalidate the cache when
-    # _latest_version changes
+    # calculate last_modified before get_old_document rolls back the document,
+    # allowing us to invalidate the cache when _latest_version changes
     last_modified = last_updated(document)
-    latest_etag = document.get(config.ETAG)
 
     # synthesize old document version(s)
     if resource_def['versioning'] is True:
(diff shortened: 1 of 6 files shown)


nckpark:
@joshvillbrandt - you'll want to review this as well.

nckpark:
First commit had ci build failures caused by an unrelated test which has the potential for random failures due to a flawed if statement. I went ahead and fixed that test to match its original intent.

nicolaiarocci:
Thanks, will wait for @joshvillbrandt feedback on this one.

nckpark:
After running into more cache issues and spending some more time with this, I discovered that Eve's cache validation behavior is inconsistent with the HTTP spec. Specifically, if a server receives both `If-Modified-Since` and `If-None-Match` cache validator headers, it should only respond with 304 if both validators pass. (see [section 13.3.4](http://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec13.3.4))

That change also makes part of my initial change - using the latest version's etag when checking cache validity - no longer necessary. The check against `last_modified` will fail, and the cache will be invalidated as expected.

And a quick heads up - I'm going to commit one more cache fix to this branch shortly to handle invalidating the cache when embedded documents have been updated.

nckpark:
I was super confused by the test failure after my last commit, as I could not reproduce it in my dev environment. After some investigation, I found that the failure was caused by not enough time passing between the creation of the root document and editing the child document for the child's last updated timestamp to invalidate the cache (`If-Modified-Since` headers only have a resolution of 1s) That is fixed in my last commit.

The more interesting discovery has to do with why I couldn't reproduce it. The TestBase random data generators set document timestamps using `datetime.now()`, while eve methods use `datetime.utcnow()`. On my dev system, the root document timestamps were set with the local time (pacific) returned by `datetime.now()`, which appeared to the cache comparison to be hours before the timestamp set while patching the child doc. On TravisCI, though, the local time was UTC, so the timestamps matched and the cache was not invalidated.

To prevent confusion like this in the future, I've updated the data generators to use date formats that match those used in the eve methods.

eve/methods/get.py
```diff
@@ -217,24 +221,32 @@ def getitem(resource, **lookup):
...
+            if embedded_last_updated > last_modified:
+                last_modified = embedded_last_updated
 
-    if req.if_modified_since and last_modified <= req.if_modified_since:
-        # request If-Modified-Since conditional request match. We test
-        # this after the etag since Last-Modified dates have lower
-        # resolution (1 second).
-        return {}, last_modified, document.get(config.ETAG), 304
+    # facilitate client caching by returning a 304 when appropriate
+    cache_validators = {True: 0, False: 0}
```
nicolaiarocci: Brilliant!

nckpark:
Last commit is my final planned change. After discovering Chrome will on occasion stop including If-Last-Modified headers and only send an If-None-Match cache validator (for reasons I cannot determine), I realized that we should not trust etag based validation for old document versions. Old etags cannot provide accurate information about the state of the client's cached _latest_version field, and therefore can result in invalid `Not Modified` responses if used as the only validator. Instead, we should ignore `If-None-Match` on old versions, and only use the `If-Modified-Since` validator.

Regarding the failed TravisCI build: Failure is another intermittent failure in eve.tests.methods.ratelimit.TestRateLimit.test_ratelimit_resource. I don't have bandwidth to debug why that test has inconsistent results, but I've observed it failing about 1 / 10 times with and without my changes.

nicolaiarocci:
> Regarding the failed TravisCI build: Failure is another intermittent failure in eve.tests.methods.ratelimit.TestRateLimit.test_ratelimit_resource. I don't have bandwidth to debug why that test has inconsistent results, but I've observed it failing about 1 / 10 times with and without my changes.

Yeah we have been having a heisenbug there for quite a while and it is tremendously irritating.

nicolaiarocci:
I went ahead and merged this with https://github.com/nicolaiarocci/eve/commit/fddb038299e5e29097150f1075ee5e55022b14c9

Thanks!

PR #614, opened 2015-04-23
Fix serialization for lists of lists

nckpark:
Attn: @joshvillbrandt


nicolaiarocci:
Thanks!

joshvillbrandt:
@nckpark mentioned that in the future, this should probably be refactored as a recursive function. While valid Cerberus schema, I think something like a list of list of lists still would fail here.

PR #616, opened 2015-04-27
condition symbol

xgdgsc:
2 equal symbol in page http://python-eve.org/tutorials/account_management.html are wrong


nicolaiarocci:
Thanks. If you want to replace your nickname with the actual name feel free to PR it.

Issue #608, opened 2015-04-20
Specific resource configuration to create Mongo indexes


pfreixes:
I've written a tentative commit [1] to get support for declare Mongo indexes as config of the resource and create these before the Eve application is launched. The idea is get together the schema definition and the index definition belonging to one resource. 

Next snippet shows one example of a new schema with one index defined:

```python
DOMAIN = {
    'resource': {
        'schema': {
            'name': {'type': 'string'}
        },
        'mongo_indexes': {
            'name': [('name', 1)]
        }
    }
}
```
The new `mongo_indexes` supports any kind of index already supported by [pymongo](http://api.mongodb.org/python/current/api/pymongo/collection.html) and also it supports compound indexes using various elements inside the list.

About the solution proposed. Current implementation of the data layer of Mongo, because of `pymongo_flask` extension, is highly coupled with the flask request life style and in this scenario there are two ways to figure it out. 

The first one is call at each request the properly `ensure_index` method - same for `create_index` - and get help from the already cache implemented by PyMongo driver to avoid call at last the creation function at Mongo service.

The second one, that was implemented, calls the `create_index` method before the Flask app is started,  where IMHO the index creation should be created. For this solution the current implementation of the Mongo data layer is not compatible. To overcome this situation I used the `pymongo` module and also I get some resource configuration options without use the `eve.config` module.

Feel free to comment anything about this feature and of course about the proposed implementation.

[1] https://github.com/pfreixes/eve/commit/eaacf3c059c1986741c0abe6dcf0d40c57cafaed

nicolaiarocci:
I've always been reluctant to handle indexes at the app layer because I feel that these should be handled very carefully at (db) design stage. However, I admit that my idea was to build some sort of auto-indexing engine while your approach is very declarative, which leaves the initiative in maintainer's hands, which is cool.

On implementation: opening a temporary connection is going to be slow performance-wise, but since you provide a static method it should not be that bad. I do realise that index creation is only going to happen at startup and before the API is made available to clients, so performance might not be very relevant at this stage. 

I would experiment with moving index creation inside `register_resource` (or better, invoke creation from there), so when people register resources after startup (as we sometime do in the test suite) they still get their indexes dealt with.

Finally, I don't think you are dealing with index options at this time, and this is going to be a issue I believe. As soon as we provide an index creation option people will want to do whatever they want with them: sparse, unique, TTL, etc.

pfreixes:
[...]

PR #618, opened 2015-05-01
#608 Get support for declare and create mongodb indexes

pfreixes:
Added a new config keyword called `mongo_indexes` that takes the list of keys that
belong to the index and also as optional the specific arguments of the index.


nicolaiarocci:
Thanks, this has been merged in https://github.com/nicolaiarocci/eve/commit/e615518768a45e7c7de95c40e341d8275b920f43. I only refactored the docs a little bit.

PR #622, opened 2015-05-08
fix non serializable attributes in document_etag()

kiniou:
Hello,

This is an attempt to fix non serializable attributes (eg. timedelta)
even if we provide a custom json encoder to the underlying data driver (see RedTurtle/eve-sqlalchemy#37 for original issue).

Actually, I am not sure this is the best way to fix this issue since it needs to get the Flask app_context() but I did not find a cleaner way.

Also, I don't have any clue about the differences of generated etag by using `json_util.default` or the `json_encoder_class.default`.

PS: I was not sure on which branch i should have based my changes so I took the most recent branch 0.5.4.


kiniou:
I'm closing this pull request since I overlooked the contributing guidelines which specify to base my pull request on the `develop` branch.

PR #624, opened 2015-05-09
fix non serializable attributes in document_etag()

kiniou:
Hello,

This is an attempt to fix non serializable attributes (eg. timedelta)
even if we provide a custom json encoder to the underlying data driver (see RedTurtle/eve-sqlalchemy#37 for original issue).

Actually, I am not sure this is the best way to fix this issue since it needs to get the Flask app_context() but I did not find a cleaner way.

Also, I don't have any clue about the differences of generated etag by using `bson.json_util.default` or the `json_encoder_class.default`.

### DIFF
--- a/eve/utils.py
+++ b/eve/utils.py
@@ -297,7 +297,11 @@ def document_etag(value, ignore_fields=None):
         value_ = value
 
     h = hashlib.sha1()
-    h.update(dumps(value_, sort_keys=True).encode('utf-8'))
+    with app.app_context():
+        json_encoder = app.data.json_encoder_class()
+        h.update(
+            dumps(value, sort_keys=True,
+                  default=json_encoder.default).encode('utf-8'))
     return h.hexdigest()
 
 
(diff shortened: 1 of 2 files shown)


eve/utils.py
```diff
@@ -297,7 +297,11 @@ def filter_ignore_fields(d, fields):
         value_ = value
 
     h = hashlib.sha1()
-    h.update(dumps(value_, sort_keys=True).encode('utf-8'))
+    with app.app_context():
```
nicolaiarocci: you probably don't need tie app_context here since it's already there when this function is invoked from the test suite.

nicolaiarocci:
Looks good, I'll fix the app_context() usage in utils.py for you.

nicolaiarocci:
Thanks! This has been merged in https://github.com/nicolaiarocci/eve/commit/72036276f2204d59e54a8a9fda5a6cc98f21143a

kiniou:
Glad to have helped ! :wink:

Issue #335, opened 2014-05-13
DELETE on a versioned item should (probably optionally) not delete archived versions.


mikenerone:
In my app, the reason I'm wanting to keep versioned documents in the first place is for auditing purposes. Without going into too much detail, events occur in our system, and we process those events according to a constantly-evolving set of rules that are defined in a MongoDB collection that is managed via an Eve API (if I have my wish). "Evolving" means that individual rules are frequently updated but also occasionally DELETED. When a rule is applied to an event, we record it (e.g. "Rule X, version N, was applied.") Later, if anything erroneous is found to have happened, we can go look at exactly how the rule was defined at the time. This currently breaks with DELETE, though, because deletion results in removal of the archived versions of the document, as well.

The workaround I currently have in mind is to basically have a `deleted` boolean field, and filter them out by default for that resource. This is clunky because it means that deletion would be done with PATCH/PUT and it just leaves the primary collection cluttered. IMO, there needs to be a way to support DELETE on the resource items while still preserving the auditability.

The behavior I would expect is that if item X is deleted, `/resource/X` would result in a 404, but that `/resource/X?version=N` would still work. Clearly, though, there's plenty of room for interpretation as to what the right semantics would be, so I haven't done any work on this. I'm just throwing it up here for consideration for the time being.

nicolaiarocci:
#626 is a pull request by @nckpark that tries to address concerns expressed in this thread. It's big and it's feature rich so I would appreciate your feedback in the PR thread, if you are still interested in this feature. Thanks!

PR #626, opened 2015-05-13
Feat/soft delete

nckpark:
Implements soft delete with consideration for the outline developed in issue #335. PR is large, but documentation in features.rst is a good place to start reviewing.

There is one noteworthy change from the implementation discussed in #335 that I'd like feedback on. I originally implemented responding to GET requests for soft deleted documents with ``410 Gone`` as described in the issue. After some more consideration and a side conversation with @joshvillbrandt, however, I moved to ``404 Not Found`` responses which include the document body. Returning the document in any 4xx response is odd, but the benefit of using 404s is that clients do not need any knowledge of soft delete to use the API correctly, whereas they would have to specially handle 410s regardless of whether or not they care about soft delete. This is especially beneficial for introducing soft delete to existing APIs.

Also of note - during testing of soft delete's behavior with embedded documents, I discovered that embedded versioned documents did not function as expected. They are fixed in commit [b928907](https://github.com/nicolaiarocci/eve/commit/b928907a87f940808e2f2e5ada32016ab2b5025a)

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -605,6 +605,10 @@ uppercase.
                                     has been deleted when ``SOFT_DELETE``
                                     is enabled. Defaults to ``_deleted``.
 
+``SHOW_DELETED_PARAM``              The URL query parameter used to include
+                                    soft deleted items in resource level GET
+                                    responses. Defaults to 'show_deleted'.
+
 =================================== =========================================
 
 .. _domain:
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -597,6 +597,14 @@ uppercase.
                                     wrapped in a ``funcname`` call. Defaults to
                                     ``None``.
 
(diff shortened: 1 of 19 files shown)


joshvillbrandt:
Looks good to me! :+1: TravisCI failure is the recurring rate limit test case bug.

CC @jenmontes for the 410->404 change.

Thanks Nick!

nicolaiarocci:
Thanks. Tests are all green now. Will review the code asap; docs seems very well done.

jenmontes:
Returning the document with the `404` will provide the necessary signal to apps aware of soft delete, but give unaware apps a familiar status code, so I think it's an excellent solution.

Fantastic work!

nicolaiarocci:
Thank you, this has been merged to 0.6-dev. Stellar work.

Issue #632, opened 2015-05-25
Remove dependency on the Flask request in Mongo datalayer


amagdas:
The current implementation for Mongo datalayer relies on the Flask request object being around in order to resolve the resource name: https://github.com/nicolaiarocci/eve/blob/develop/eve/io/mongo/mongo.py#L771
This has the obvious side effect that this is not reusable outside of a request context which is not a problem for a regular Eve app but it is if you want to actually reuse the business logic declared in the resource event handlers from a Celery worker for example.
As noted in the comment, passing the resource name down the stack is going to break the backwards compatibility for the MediaStorage but I think this is worthwhile.
Please let me know if it's ok to send a PR with these changes.

nicolaiarocci:
I took the "let's not break things" approach there, and I knew it would cause concerns. In theory since we're still beta we're allowed to break backward compatibility and in fact we've been doing that in occasions. But people are using this thing in production everywhere so I'm hesitant. 

One approach you could take with you PR is add the parameter with a None default. If it's None then fallback to the `request` object call. This would probably not break things and still allow for an evolution of the interface. We can also make it clear both in docstrings and changelog that in a future release the dependency on Flask request object will be removed from the data layer.

rs:
+1 for breaking backward compat in a 0.6 or 0.7 in exchange to an flask / data layer decoupling.

amagdas:
Today or tomorrow I will create the PR for this.

nckpark:
Also +1 for this change.

amagdas:
Created a PR for this, feedback please.

PR #635, opened 2015-05-27
Remove dependency on the Flask request in Mongo datalayer

amagdas:
Fix for https://github.com/nicolaiarocci/eve/issues/632
The only existing dependency is on the app.config, this shouldn't be a problem in most of the cases.
For Celery, we just add the Eve app config to Celery app config.

### DIFF
--- a/eve/io/media.py
+++ b/eve/io/media.py
@@ -27,14 +27,14 @@ class MediaStorage(object):
         """
         self.app = app
 
-    def get(self, id_or_filename):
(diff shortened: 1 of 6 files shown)


amagdas:
Not sure why the build is failing, it runs fine locally.

nicolaiarocci:
Test is green now. Code looks ok to me.

nicolaiarocci:
Thanks!

Issue #636, opened 2015-05-27
Projection allowing mixing inclusion and exclusion fields


slamuu:
Mongo does not allow mixing inclusion and exclusion in projection [[1](http://docs.mongodb.org/manual/tutorial/project-fields-from-query-results/)]

However, develop branch in [eve/flaskapp.py](https://github.com/nicolaiarocci/eve/blob/develop/eve/flaskapp.py#L574) projections are allowing a mix of inclusions and exclusions

Example [gist](https://gist.github.com/slamuu/841783398ba577a57cd6) with field projection

Resulting error:
```
OperationFailure: database error: Can't canonicalize query: BadValue Projection cannot have a mix of inclusion and exclusion.
```

amagdas:
This is quite a regression, just tried to upgrade our app to develop branch, basically all the exclusion projections are not working anymore.

nicolaiarocci:
Yup this has been introduced in a9d65bed. @nckpark can you look into it? Thanks.

amagdas:
Is anyone working on this one? If not, I can give it a try tomorrow.

hbarroso:
I'm not. I tried but I got into an issue regarding the SOFT_DELETE

nckpark:
Thanks for tackling this @amagdas. I've been swamped the past week and didn't get a chance to look at it. Sorry for the regression.

PR #638, opened 2015-05-28
Fix for issue #636

hbarroso:


amagdas:
It would be nice to include a test for this so we will be able to catch possible regressions.

Issue #639, opened 2015-05-28
Handling custom MediaStorage files urls


hbarroso:
Hey guys,

So, I'm currently implementing an custom MediaStorage to be used with S3. Everything works fines, however I really don't like how Eve handles the returned media file urls. 

When I use ```RETURN_MEDIA_AS_URL``` this always returns the URI related to the API address, but I actually needed to return a full URL like: ```"https://s3-us-west-2.amazonaws.com/foo/<:file_id>"``` but I can't because the ```resolve_media_files()``` method in eve.methods.common does't allow it.

It makes perfect sense for GridFS but custom MediaStorage not so much

```
...
        if _file:
            # otherwise we have a valid file and should send extended response
            # start with the basic file object
            if config.RETURN_MEDIA_AS_BASE64_STRING:
                ret_file = base64.encodestring(_file.read())
            elif config.RETURN_MEDIA_AS_URL:
                ret_file = '%s/%s/%s' % (app.api_prefix,
                                         config.MEDIA_ENDPOINT,
                                         file_id)
            else:
                ret_file = None
...
```

I think that ret_file formatted string should be user customized  if needed. Is there a way around this for the time being ?

Thanks :)

hbarroso:
I just sent a pull_request that allows this here: #640 . Let me know if this would be a good solution. Thanks

nicolaiarocci:
See my comment on the PR thread. Thanks!

PR #640, opened 2015-05-28
Added 'MEDIA_CUSTOM_URL' setting for #639

hbarroso:
Allows media urls to use a custom URL as a base pattern.
When 'RETURN_MEDIA_AS_URL' is set to True and 'MEDIA_CUSTOM_URL' is not None, any media files
in a collection will return a concatenation of this setting plus the file ID.

This allows custom 'MediaStorage' to be much more flexible.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -41,6 +41,7 @@ Patches and Contributions
 - Grisha K.
 - Hannes Tiede
 - Harro van der Klauw
+- Henrique Barroso
 - James Stewart
 - Jaroslav Semančík
 - Javier Gonel
--- a/eve/default_settings.py
+++ b/eve/default_settings.py
@@ -174,7 +174,7 @@ RETURN_MEDIA_AS_BASE64_STRING = True
 RETURN_MEDIA_AS_URL = False
 MEDIA_ENDPOINT = 'media'
 MEDIA_URL = 'regex("[a-f0-9]{24}")'
-MEDIA_CUSTOM_URL = None
+MEDIA_BASE_URL = None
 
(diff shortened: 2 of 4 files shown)


nicolaiarocci:
Hello this looks like a nice addition. 

Perhaps you could rename MEDIA_CUSTOM_URL to `MEDIA_BASE_URL` and allow  `MEDIA_ENDPOINT` to be used as well. 

Possible implementation:

    prefix = config.MEDIA_BASE_URL if config.MEDIA_BASE_URL is not None else app.api_prefix
    ret_file = '%s/%s/%s' % (prefix, config.MEDIA_ENDPOINT, file_id)

So in your case you would set things up like this:

    MEDIA_BASE_URL = 'https://s3-us-west-2.amazonaws.com'
    MEDIA_ENDPOINT = 'foo'

This implementations seems more in line with how things work when only MEDIA_ENPOINT is used. What do you think?

PS: I would be happy to add your S3 implementation on Eve extensions page.

hbarroso:
@nicolaiarocci thanks for the feedback, I will update my PR to match your code. It looks a lot cleaner than mine.

As for the S3 extension, I still need to package it as 'module' currently it's just a single class file. From next week I might have something to share :)

nicolaiarocci:
Looking forward it!

nicolaiarocci:
Thanks!

hbarroso:
No problem, I'm glad I could contribute to Eve :)

nicolaiarocci:
I also updated the docs with a working example. See http://python-eve.org/features#serving-media-files-at-a-dedicated-endpoint

PR #641, opened 2015-05-31
Do not overwrite the _id field as it is not a sub resource

rs:
This PR fix an issue with `post_internal` when the `_id` field is specified in the posted document and the current request has its own `_id` parameter. The `_id` field of the posted document get replaced by the `_id` parameter of the current request.

This fix does only hide the real issue though. The use of `request.view_args` in `resolve_sub_resource_path` doesn't play well with `*_internal` methods as it tries to alter the document with data coming from the current request which may have no link with the `*_internal` request.

I guess a stronger fix would be to pass the route's parameters as an argument to the `*_internal` methods.

### DIFF
--- a/eve/methods/common.py
+++ b/eve/methods/common.py
@@ -808,7 +808,7 @@ def resolve_sub_resource_path(document, resource):
     schema = app.config['DOMAIN'][resource]['schema']
     fields = []
     for field, value in request.view_args.items():
-        if field in schema:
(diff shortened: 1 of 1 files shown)


nicolaiarocci:
Merged, although it would be good to cover this fix with its own test. Long term I agree we should continue in our effort to decouple the data layer from flask objects (see #635).

PR #642, opened 2015-06-01
Added option to use a different authorization database.

Expi1:
Hi,

I added a new configuration value, `MONGO_AUTHDBNAME`, that is passed to [PyMongo's authenticate function](http://api.mongodb.org/python/current/examples/authentication.html#delegated-authentication) and should allow for authentication on databases different from the main database that's used for the data storage.

I've added it to the documentation, but unfortunately was unable to find where it should be included in the tests, however, all the tests that are existing, run fine with no issues. I'd appreciate some pointers to where tests should be included for this.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -102,3 +102,4 @@ Patches and Contributions
 - xgdgsc
 - boosh
 - dccrazyboy
+- David Wood
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -477,6 +477,8 @@ uppercase.
 
 ``MONGO_DBNAME``                    MongoDB database name.
 
+``MONGO_AUTHDBNAME``                MongoDB authorization database name. Defaults to ``None``.
+
 ``MONGO_MAX_POOL_SIZE``             The maximum number of idle connections 
                                     maintained in the PyMongo connection pool.
                                     Default: PyMongo default.
--- a/eve/io/mongo/mongo.py
+++ b/eve/io/mongo/mongo.py
@@ -872,6 +872,8 @@ def create_index(app, resource, name, list_of_keys, index_options):
         if key('USERNAME') in app.config else None
     password = app.config[key('PASSWORD')] \
         if key('PASSWORD') in app.config else None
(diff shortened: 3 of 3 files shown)


nicolaiarocci:
Hello and thanks for your contribution. 

Delegated Authentication was introduced in MongoDB 2.4 and [removed on 2.6](http://api.mongodb.org/python/current/examples/authentication.html#delegated-authentication). With the advent of MongoDB 3.0 I wonder if we want to add support for this feature, given that it has been abandoned for a while. 

Thoughts?

Expi1:
Hi,

I never noticed that feature had been removed. 

I made this change because I had been creating a docker image for an API using Eve, and was attempting to link it with the `tutum/mongodb` image by [the folks over at tutum.co](http://tutum.co). I was getting an authentication error, when I asked the developers behind the `tutum/mongodb` image I was informed that I'd have to authenticate against the `admin` database first. 

After some research, I made this change to allow for alternate authentication databases, however, I might be referring to the wrong feature when I mention Delegated Authentication - because I'm certain that the database image I'm using is version 3.0 of MongoDB, and requires me to auth on a different database. I wonder if [this mongo shell parameter](http://docs.mongodb.org/manual/reference/program/mongorestore/#cmdoption--authenticationDatabase) refers to the removed Delegated Authentication feature or something else?

nicolaiarocci:
That link is about `mongorestore` but yes, even the [Mongo 3.0 docs](http://docs.mongodb.org/manual/reference/program/mongo/#cmdoption--authenticationDatabase) don't mention the feature as being deprecated/retired since 2.6. Also, PyMongo [changelog](http://api.mongodb.org/python/current/changelog.html) does not seem to report any significant change to authentication delegation since PyMongo 2.5 which introduced it.

Let's assume that's an error in PyMongo documentation. When I get back I will see if I can somehow add a test for this. Stay tuned.

Expi1:
Hi,

Any update on this?

nicolaiarocci:
Merged with https://github.com/nicolaiarocci/eve/commit/235c66f488f6b4579bc2b473187673913cd80682

Thanks!

PR #643, opened 2015-06-02
Improve performance of GET requests to versioned resources

nckpark:
Synthesizing a specific document version in GET requests uses two `deepcopy` calls, but only one is necessary if `get_old_document` does not edit the document passed to it. Improves performance for large documents.


nicolaiarocci:
Thanks

PR #647, opened 2015-06-05
Proper fix for exclusion projection. Fix for #636

amagdas:


eve/flaskapp.py
```diff
@@ -560,10 +560,15 @@ def _set_resource_defaults(self, resource, settings):
         settings['datasource'].setdefault('filter', None)
         settings['datasource'].setdefault('default_sort', None)
 
-        projection = settings['datasource'].get('projection')
+        projection = settings['datasource'].get('projection', {})
         projection = projection or {}
```
nicolaiarocci: Superfluous?

amagdas: Nop. Build failed without it.

PR #649, opened 2015-06-08
Only initialize oplog resource if oplog is enabled. (Closes #648)

### LINKED ISSUE #648: Oplog resource should not be initialized if oplog is disabled
Commit [1476107](https://github.com/nicolaiarocci/eve/commit/1476107fb99a3e88beaff6e289682958e5f59f15) modified `_init_oplog` behavior to always initialize the oplog resource. This resource should only be created if the `OPLOG` config option is set to True.

nckpark:


nckpark:
TravisCI failures from first commit revealed an intermittent failure in the `tests/versioning.py` `test_softdelete` test, caused by the test not expecting diffs to include the `_links` field which does change with a new version. Because the test accounts for the possibility that the speed of test execution will sometimes leave `last_updated` unchanged, meaning one extra field changes, the unexpected `_links` field did not cause the test to fail in the majority of runs. Fixed in [e3cb993](https://github.com/nicolaiarocci/eve/commit/e3cb99380fe32059c45bf45bf7c9e8e40fb7648e).

nckpark:
Remaining TravisCI failure is the known intermittent `test_ratelimit_resource` failure.

PR #651, opened 2015-06-08
Update serialize to gracefully handle non-dictionary values in dict type fields.

nckpark:
Nullable fields of type `dict` are not currently possible, as `methods.common.serialize` breaks when attempting to recursively serialize a null value. The same is true of other non-dictionary value types, which should fail in validation, not raise an exception during serialization. This PR allows serialization to continue, nullable `dict` fields to work as expected, and type errors to be reported accurately from the validator.


nicolaiarocci:
Merged in https://github.com/nicolaiarocci/eve/commit/82984139f0e903fc011e529ef4740d4374b32de8

PR #655, opened 2015-06-12
Using constant VERSION instead of hard coded '_version'

mdhaman:
GET fails if you change the _version to some other field name.


nicolaiarocci:
Thanks!

PR #656, opened 2015-06-12
embedded documents following depth level

goneri:
embedded_fields may point on a field that come from another embedded
document. For example, ['a.b.c', 'a.b', 'a']

The strategy is to sort out the different embedded_fields depending on
their number of '.'. This way, the ones with the greatest depth level
come last.

Combined with https://github.com/RedTurtle/eve-sqlalchemy/pull/47 it's
now possible to do include multi-level embedded resource with Eve-SQLAlchemy.

### DIFF
--- a/eve/methods/common.py
+++ b/eve/methods/common.py
@@ -601,7 +601,8 @@ def resolve_embedded_documents(document, resource, embedded_fields):
 
     .. versionadded:: 0.1.0
     """
-    for field in embedded_fields:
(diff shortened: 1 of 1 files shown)


nicolaiarocci:
Could you please re-submit against `develop`, and rebase onto the top of that branch? We don't merge to `master`. Thanks.

goneri:
against “develop” this time→ https://github.com/nicolaiarocci/eve/pull/664

PR #658, opened 2015-06-18
WWW-Authenticate header syntax fix

stt:
According to [rfc2617](http://tools.ietf.org/html/rfc2617#section-2) the separator should be (=) instead of (:).
This caused at least chrome not to prompt user for the credentials, and not to send the Authorization header even when credentials were in the url.


nicolaiarocci:
Excellent, thank you!

Merged in https://github.com/nicolaiarocci/eve/commit/74ddb50f4500db86e6758d003b9b0b5ecee32218

PR #662, opened 2015-06-23
Add some tests to demonstrate how dependencies with values doesn't work with patch and put

rs:
I'm trying to use new Cerberus' `dependencies` with value deps, and I can't make it to fail when the dependency is not met.

I've pushed some failing test to demonstrate my issue. Did I miss something?

### DIFF
--- a/eve/tests/methods/patch.py
+++ b/eve/tests/methods/patch.py
@@ -552,6 +552,33 @@ class TestPatch(TestBase):
                                headers=[('If-Match', etag)])
         self.assert200(status)
 
+    def test_patch_dependent_field_value_on_origin_document(self):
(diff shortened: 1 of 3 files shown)


nicolaiarocci:
At quick glance I would say that since `dependency_field1` has a `default` value, validation passes on `put` since the dependent field value will be injected. Patch is another story though. I'm on it.

rs:
The default value is `default` while the dependent value is `value`, so event with default it should return a `422`.

nicolaiarocci:
Right.

rs:
Awesome thx!

BTW, when do you think 0.5.4 will come out?

nicolaiarocci:
Actually I'm going straight for 0.6. Should be soon.

PR #665, opened 2015-07-02
Dependency doesn't play well with required

rs:
If a field has both a dependency and a required directive, the required directive is just ignored. This PR contains a failing test to demonstrate the issue.

### DIFF
--- a/eve/tests/methods/post.py
+++ b/eve/tests/methods/post.py
@@ -482,6 +482,13 @@ class TestPost(TestBase):
         data = {test_field: test_value}
         self.assertPostItem(data, test_field, test_value)
 
+    def test_post_dependency_required_fields(self):
+        del(self.domain['contacts']['schema']['ref']['required'])
+        self.domain['contacts']['schema']['dependency_field3']['required'] = True
+        r, status = self.post(self.known_resource_url, data={})
+        self.assertValidationErrorStatus(status)
+        self.assertValidationError(r, {'dependency_field3': 'required'})
+
     def test_post_dependency_fields_with_values(self):
         # test that dependencies values are validated correctly. See #547.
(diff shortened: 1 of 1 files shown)


nicolaiarocci:
That's actually how it is supposed to work right now. [If a required field also has dependencies and it is missing, then we only report it as missing if the dependencies validated](https://github.com/nicolaiarocci/cerberus/blob/master/cerberus/cerberus.py#L471-L476).

rs:
In my case the dependency is met thru the default value of the depending field but still doesn’t complain if the required field is omitted. I can change the test to meet that case if you want.

Also I think that when you use required, you expect that whatever the other directives, you won’t end up with a missing field in the db.

nicolaiarocci:
> Also I think that when you use required, you expect that whatever the other directives, you won’t end up with a missing field in the db.

it used to be like this. New behaviour was introduced with nicolaiarocci/cerberus/pull/61 by @hthieu1110. [This test](https://github.com/hthieu1110/cerberus/blob/2ab5ae4984e644f0a34a548400eac74d04753f5a/cerberus/tests/tests.py#L505-L507) in particular shows that this is in fact the intended behaviour. 

Now, I agree that this is somewhat convoluted. `required` being validated only when dependencies are met seems counter-intuitive. In this scenario it seems appropriate to give precedence to the `required` rule over `dependencies`.

I'd appreciate if @hthieu1110, as the original patch author, would chime in on this one.

gmauleon:
Sorry to jump in so late, just starting to use Eve. But then how could I require a field only if another field is of value X?

I got a field "mytype" with 2 possible value A and B, I want that when "mytype" is A only, then field "myfield" is required. I did that with a dependency set on the field "myfield" that need "mytype" to be A. But this patch broke that logic. Any workaround ?

Thanks!

PR #670, opened 2015-07-14
Fix a crash when no Content-Type Header is provided

sebest:


nicolaiarocci:
Thanks for this. Would be nice to also have test coverage for this so we don't have regressions in the future. Would you add it?

sebest:
sure, any recommendation on where (which file) i should put this test in and if i should create a new test class or use an existing one, when i checked i did not find where this function is tested so i conclude it is not tested yet?

nicolaiarocci:
Actually, I've been trying to reproduce this with no luck. Flask defaults to an empty string `Content-Type` when it's not provided, which does not break the code and gracefully returns a 400 with proper 'CT missing' message. I tried this with Curl. Can you let me know how you got the hard crash?

Of course the code you suggest is still better than the original (it adds a guard against future Flask regressions), so I'm willing to merge anyways.

sebest:
I don't remember exactly how i triggered this bug, but i sure did.
I think we should merge the code as it is safer anyway.

nicolaiarocci:
Thanks. Merged with https://github.com/nicolaiarocci/eve/commit/ee7410a8932aa2cefb3d455cac823dd3d7d9f919

PR #672, opened 2015-07-15
Log Exception using app.logger.exception()

sebest:


nicolaiarocci:
Thanks for this. Indeed internal app logging is needed, and it is on the roadmap probably around v0.7. I'm putting this on on hold for the time being as I first want to figure out the general design for logging (i.e. we probably want a single logging function, accessible by all exception handles *and* user functions, etc.)

rs:
Could we have at least that in the meantime? Because as of today, it’s impossible to figure out bugs happening in production or even debug a hook in dev without editing eve’s code.

sebest:
Yeah i think that even if we improve logging for 0.7 having a log for base Exception is mandatory today.

nicolaiarocci:
See http://python-eve.org/features#enhanced-logging

rs:
👍

PR #675, opened 2015-07-23
Fix versioning of resources which define a custom ID_FIELD

matthieuprat:
When versioning is enabled on a resource with a custom `ID_FIELD`, versioning documents will inherit their ID from the versioned document, making any update of the document result in a `DuplicateKeyError`.

This PR adresses this issue by excluding the `ID_FIELD` from the list of fields returned by `eve.versioning.versioned_fields`.


nicolaiarocci:
Thanks. Merged with https://github.com/nicolaiarocci/eve/commit/d3a2faa16951c1fdd694185eae9f6f74f1ccda88

PR #676, opened 2015-07-28
Removing some calls to debug_error_message to return the real error m…

sebest:
…essage in API response.


nicolaiarocci:
Thanks. Merged https://github.com/nicolaiarocci/eve/commit/bd7b12c7b46a240ea78a9ca63cca96c4a1de4c7e

PR #683, opened 2015-08-13
Added PUT to item_methods config docs

stratosgear:
I was wondering where was the PUT verb in your Resources or Item methods, until I tried some code samples and I just saw that you correctly accept PUT in item_methods but the docs do not accurately reflect that.


nicolaiarocci:
Good catch, thanks!

Issue #685, opened 2015-08-17
Provide a basic home page response when HATEOAS is off


stratosgear:
What would you think about returning some basic EVE info in the home page?

An example, from elasticsearch, to illustrate my point:

    {
        status: 200,
        name: "Julie Power",
        cluster_name: "elasticsearch",
        version: {
            number: "1.7.1",
            build_hash: "b88f43fc40b0bcd7f173a1f9ee2e97816de80b19",
            build_timestamp: "2015-07-29T09:54:16Z",
            build_snapshot: false,
            lucene_version: "4.10.4"
        },
        tagline: "You Know, for Search"
    }

Not all of the fields makes sense for Eve, I know, but I think you get my point.

This can also serve as a nice way to check if Eve is still up an running if you are pinging your server to make sure it is still up and running and responding to requests.

nicolaiarocci:
When HATEOAS is disabled and you hit the homepage:

- You can check the `Server` header to make sure that Eve us up and running (it will contain something like `Eve/0.6 Werkzeug/0.10.4 Python/2.7.8`);
- if debug mode is on, you get a `HATEOAS is disabled so we have no data to display at the API homepage` message.

Now of course you don't use debug mode in production, so I guess you are left with the server header option. Would that be enough for both pinging the API and checking its version? Since HATEOAS has been disabled, I can't think of any other information might be useful there.

stratosgear:
Some more comments...

Hitting the home page (with HATEOAS off) it results in:

    {
      "validation": "ERR",
      "error": {
        "message": "The requested URL was not found on the server.  If you entered the URL manually please check your spelling and try again.",
        "code": 404
      }
    }

This is, imho, an "ugly" page.  I can read and understand the error message, but it is not a friendly welcome page as the home page to a new user of the API.  It looks like something is not working right.

On the other hand, a status/version reply (as in the elasticsearch example above) looks more "comforting" that at least everything is working as intended and gives an easy to read info about the server, without having to investigate headers (which are missing the API version, btw).

So all in all, I consider the home page as the first page someone will try to test, and seeing an error page gives a wrong/unprofessional impression, especially when something can be easily implemented and to provide some useful info to the end users.

But that's just **my** take on the matter... :)

stratosgear:
Just created a PR that implements what I mentioned above (with tests too)

It replies with a simple:

    { 'eve_version': 'a.b.c', 'api_version': 'x.y.z'}

[...]

PR #686, opened 2015-08-18
Responds with a valid response when HATEOAS is off

stratosgear:
When HATEOAS is turned off, the server now responds with
an json response, stating the EVE and API versions, instead
of replying with a 404 error page.

Putting my money where my mouth is, regarding #685

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -406,6 +406,14 @@ uppercase.
 
 ``META``                            Allows to customize the meta field. Defaults
                                     to ``_meta``
+                                    to ``_meta``.
+
+``INFO``                            String value to include an info section, with the
+                                    given INFO name, at the Eve homepage (suggested
+                                    value ``_info``). The info section will include
+                                    Eve server version and API version (API_VERSION,
+                                    if set).  ``None`` otherwise, if you do not want
+                                    to expose any server info. Defaults to ``None``.
 
 ``LINKS``                           Allows to customize the links field. Defaults
                                     to ``_links``.
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -391,12 +391,6 @@ want to turn HATEOAS off? Well, if you know that your client application is not
 going to use the feature, then you might want to save on both bandwidth and
 performance. 
 
-.. admonition:: Please note
-
-    When HATEOAS is disabled, the API entry point (the home page) will return
-    a ``404``, since its only usefulness would be to return a list of available
-    resources, which is the standard behavior when HATEOAS is enabled.
-
 .. _jsonxml:
 
 JSON and XML Rendering
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -57,6 +57,11 @@ payload:
 ::
 
     {
+        "_info": {
+            "server": "Eve",
+            "version": "a.b.c",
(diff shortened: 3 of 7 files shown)


nicolaiarocci:
I'm thinking that at this point we might want to provide the same metadata regardless of HATEOAS state. We might respond with something like

    {
        "_meta": {
            "server":  "Eve",
            "version": 0.6,
            "api_version": <api version>
        },
        "_links": {...}
    }

This way the home endpoint would end up always returning a `_meta` (or `_home` or `_version` maybe?) field. If HATEOAS is disabled the `_links` field will be gone. Thoughts?

stratosgear:
Counter-suggestion:

I always thought that the name ```_meta``` was kind of unfortunate since all it does is convey pager information.

No need to complain though, because you have very thoughtfully created a ```META``` config, and I have already renamed ```_meta``` to ```_pager```, in my appplication.

Continuing along this line of thought, the additional ```server```, ```version``` and ```api_version``` would "pollute" the ```_meta``` (or ```_pager``` if you want) namespace.

May I suggest an ```_info``` or ```_server_info``` (with the appropriate ```INFO``` or ```SERVER_INFO``` config variables) to host the ```server```, ```version``` and ```api_version``` values...?

nicolaiarocci:
Looks good, I like the addition the corresponding config variables. :+1:

stratosgear:
You still haven't specified which of the ```_info``` or ```_server_info``` options you like though...! :)

I vote for ```_info``` since it is shorter.  Let me know and I'll try to amend my PR...

nicolaiarocci:
What you said.

stratosgear:
I did my best and have updated the PR, but I give up... :(

I commited twice and Travis, both times, complains with some:

    FAIL: test_ratelimit_item (eve.tests.methods.ratelimit.TestRateLimit)

in unrelated test methods.  I only changed the docs between the two commits

Check Travis builds:

* https://travis-ci.org/nicolaiarocci/eve/builds/76456405
* https://travis-ci.org/nicolaiarocci/eve/builds/76509283

Seems like the build server is overloaded or something and the tests are failing randomly...? Have you seen this before?

Sorry, I might try this again some other time...

nicolaiarocci:
It's ok, it's all green now. We have a heisenbug with that test and I should really be fixing it one of these days. Usually I simply need to re-launch the travis job to get it to pass. I'll review the updated PR tomorrow. Thanks!

nicolaiarocci:
I would also allow `INFO` to be set to `None`, which would preserve the current behaviour (404 on homepage if HATEOAS is disabled; `_links` if enabled). Not everybody will probably feel good with leaking server info to the public. For the same reason I wold also make this an opt-in, setting `INFO` to `None` as default.

stratosgear:
Fair enough... wait while I amend the PR!

stratosgear:
I am still making the assumption though, that the server responds with a 200 instead of a 404.  It will not just "leak" server info, though, just an empty dict.

stratosgear:
Done from my side!

nicolaiarocci:
Merged with https://github.com/nicolaiarocci/eve/commit/4affaefb841c08067aa37488ce07b1eae047706a

PR #698, opened 2015-09-02
Fix resource discovery in Mongo layer

amagdas:
Mongo layer should not depend on request context if possible, resource can be passed as a parameter.


amagdas:
@nicolaiarocci is there a relative date when the release of Eve 0.6 will happen?

nicolaiarocci:
This month.

gmauleon:
I believe this will also fix oplog_push taking the current endpoint resource instead of the oplog one for mongo prefixes

nicolaiarocci:
thanks!

PR #701, opened 2015-09-08
Apply Cerberus' type coercion to documents

dkellner:
As far as I understood type coercion correctly, a successfully validated document should show the coerced values instead of the original ones. This commit changes Eve's behaviour regarding type coercion so coerced values get applied before the document is inserted or updated.

### DIFF
--- a/eve/methods/patch.py
+++ b/eve/methods/patch.py
@@ -159,6 +159,9 @@ def patch_internal(resource, payload=None, concurrency_check=False,
             validation = validator.validate_update(updates, object_id,
                                                    original)
         if validation:
+            # Apply coerced values
+            updates = validator.document
+
             # sneak in a shadow copy if it wasn't already there
             late_versioning_catch(original, resource)
 
--- a/eve/methods/post.py
+++ b/eve/methods/post.py
@@ -181,6 +181,9 @@ def post_internal(resource, payl=None, skip_validation=False):
             else:
                 validation = validator.validate(document)
             if validation:  # validation is successful
+                # Apply coerced values
+                document = validator.document
+
                 # Populate meta and default fields
                 document[config.LAST_UPDATED] = \
                     document[config.DATE_CREATED] = date_utc
--- a/eve/methods/put.py
+++ b/eve/methods/put.py
@@ -150,6 +150,9 @@ def put_internal(resource, payload=None, concurrency_check=False,
             validation = validator.validate_replace(document, object_id,
                                                     original)
         if validation:
+            # Apply coerced values
+            document = validator.document
+
             # sneak in a shadow copy if it wasn't already there
             late_versioning_catch(original, resource)
 
--- a/eve/tests/methods/patch.py
+++ b/eve/tests/methods/patch.py
@@ -598,6 +598,16 @@ class TestPatch(TestBase):
                                headers=headers)
         self.assert200(status)
 
+    def test_patch_type_coercion(self):
+        schema = self.domain[self.known_resource]['schema']
+        schema['aninteger']['coerce'] = lambda string: int(float(string))
+        changes = {'ref': '1234567890123456789054321', 'aninteger': '42.3'}
+        r, status = self.patch(self.item_id_url, data=changes,
+                               headers=[('If-Match', self.item_etag)])
+        self.assert200(status)
+        r, status = self.get(r['_links']['self']['href'])
+        self.assertEqual(r['aninteger'], 42)
+
     def assertPatchResponse(self, response, item_id, resource=None):
         id_field = self.domain[resource or self.known_resource]['id_field']
         self.assertTrue(STATUS in response)
--- a/eve/tests/methods/post.py
+++ b/eve/tests/methods/post.py
@@ -648,6 +648,12 @@ class TestPost(TestBase):
         self.assertPostResponse(r)
         self.assertEqual(r['_id'], id)
 
+    def test_post_type_coercion(self):
+        schema = self.domain[self.known_resource]['schema']
+        schema['aninteger']['coerce'] = lambda string: int(float(string))
+        data = {'ref': '1234567890123456789054321', 'aninteger': '42.3'}
+        self.assertPostItem(data, 'aninteger', 42)
+
     def perform_post(self, data, valid_items=[0]):
         r, status = self.post(self.known_resource_url, data=data)
         self.assert201(status)
--- a/eve/tests/methods/put.py
+++ b/eve/tests/methods/put.py
@@ -317,6 +317,16 @@ class TestPut(TestBase):
         self.assert400(status)
         self.assertTrue('immutable' in r['_error']['message'])
 
+    def test_put_type_coercion(self):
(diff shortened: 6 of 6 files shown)


dkellner:
Well, after being busy the last couple of days, I finally got some time to look into this again. The failing test for Python 2.6 is because of differing behaviour of deepcopy on BytesIO objects:

```python
from io import BytesIO
import copy

s = BytesIO('foo')
t = copy.deepcopy(s)
assert s.read() == t.read()  # fails in Python 2.6, but works in more recent versions
```

In general, I think it is not advisable to do a deepcopy on the complete document in Cerberus. It will fail for the above case and - even in more recent python versions - for documents with embedded file handles. What about just doing a shallow copy of the document and updating/replacing only coerced fields with their new values? In this case, it would be safe for all applications to just use validator.document.

I understand this is more of a Cerberus-related question, but I'd like to hear your thoughts on this before pursuing further steps.

nicolaiarocci:
Thanks for digging into this. 

I suppose [this](https://bugs.python.org/issue1515) is the issue we are dealing with. It was originally scheduled for 2.6 but then slipped up to 2.7. Just our luck :smile: 

> In general, I think it is not advisable to do a deepcopy on the complete document in Cerberus. It will fail for the above case and - even in more recent python versions - for documents with embedded file handles. 

The second scenario should be handled by falling back to a shallow copy ([code](https://github.com/nicolaiarocci/cerberus/blob/0.9.1/cerberus/cerberus.py#L262-L267)). 

> What about just doing a shallow copy of the document and updating/replacing only coerced fields with their new values? In this case, it would be safe for all applications to just use validator.document.

I tested shallow copies only approach. Both Cerberus and Eve test suites pass on all Pythons. Overall I think that your approach would be ideal given the circumstances. We will need a Cerberus 0.9.2 release though, as 0.10 has breaking changes and I don't wish Eve 0.6 to deal with those (release is imminent)

dkellner:
> The second scenario should be handled by falling back to a shallow copy (code).

For me, this does not work on Python 2.7. It does deepcopy the file handle, resulting in a closed one without throwing any exception. But as this is a Cerberus issue, I'll post more details over there.
(https://github.com/nicolaiarocci/cerberus/issues/147)

Btw, I finally got Tox with Python2.6 running on this machine, so hopefully there won't be a pull request with failing checks from my side anymore ;).

nicolaiarocci:
Could you rebase on top of current `develop`? Should fix the py26 failures. Thanks.

dkellner:
Strange. I've rebased and now a redis test is failing. On my machine this test is working (even with Python 2.6):

```
test_noratelimits (eve.tests.methods.ratelimit.TestRateLimit) ... ok
test_ratelimit_home (eve.tests.methods.ratelimit.TestRateLimit) ... ok
test_ratelimit_item (eve.tests.methods.ratelimit.TestRateLimit) ... ok
test_ratelimit_resource (eve.tests.methods.ratelimit.TestRateLimit) ... ok
----------------------------------------------------------------------
Ran 4 tests in 6.980s
OK
```

Any ideas?

nicolaiarocci:
It's ok now. We have a heisenbug there.

nicolaiarocci:
Thanks!

gwainer:
Hello, 

I am using eve 0.6.4 with python 3.5.1. I have type coercion function[1] in a boolean field similar to the examples in the Cerberus documentation. Based on @dkellner initial comment here, I was expecting that the inserted document would have the coerced values instead of the original ones. But this is not happening. 

Is this the expected behaviour? Thanks!

[1] - to_bool = lambda v: v if type(v) is bool else str(v).lower() in ['true', '1']

dkellner:
@gwainer: I just tried to reproduce your problem with a minimal application:
```python
to_bool = lambda v: v if type(v) is bool else str(v).lower() in ['true', '1']
DOMAIN = {
    'foo': {
        'schema': {
            'bool': {'type': 'boolean', 'coerce': to_bool}
        }
    }
}
```

It works for me (Eve 0.6.4 with Python3.4 as i don't have 3.5 available right now):

```
$ curl -X POST -H 'Content-Type: application/json' -d '{"bool": 1}' http://localhost:5000/foo
{"_id": "57d366c8e288f01885d5afa9", "_status": "OK", "_etag": "6a543688ec4918a544a17e639b1d321afab8ae1e", "_links": {"self": {"href": "foo/57d366c8e288f01885d5afa9", "title": "Foo"}}, "_updated": "Sat, 10 Sep 2016 01:50:00 GMT", "_created": "Sat, 10 Sep 2016 01:50:00 GMT"}

$ curl -X POST -H 'Content-Type: application/json' -d '{"bool": "false"}' http://localhost:5000/foo
{"_id": "57d366f0e288f01885d5afaa", "_status": "OK", "_etag": "189f2d351578311344f226c92e1f800c9e98f529", "_links": {"self": {"href": "foo/57d366f0e288f01885d5afaa", "title": "Foo"}}, "_updated": "Sat, 10 Sep 2016 01:50:40 GMT", "_created": "Sat, 10 Sep 2016 01:50:40 GMT"}
```

Mongo-CLI confirms that documents were stored with correct bool values:

```
> db.foo.find()
{ "_id" : ObjectId("57d366c8e288f01885d5afa9"), "_etag" : "6a543688ec4918a544a17e639b1d321afab8ae1e", "_updated" : ISODate("2016-09-10T01:50:00Z"), "bool" : true, "_created"
: ISODate("2016-09-10T01:50:00Z") }
{ "_id" : ObjectId("57d366f0e288f01885d5afaa"), "_etag" : "189f2d351578311344f226c92e1f800c9e98f529", "_updated" : ISODate("2016-09-10T01:50:40Z"), "bool" : false, "_created" : ISODate("2016-09-10T01:50:40Z") }
```

Can you confirm a minimal example like this is working in your setup? I can try it later with Python 3.5 too, but I am still travelling at the moment and therefore quite limited ;).

gwainer:
Thank you for the quick response @dkellner. I can confirm that your example works. 

I may have found a corner-case which is not working. Let me show a simplified example of schema that doesn't work. It is a list of dicts with the boolean inside the dicts:

```javascript
DOMAIN = {
    'foo': {
        'entries': {
            'type': 'list',
            'schema': {
                'type': 'dict',
                'schema': {
                    'time': {
                        'type': 'string'
                    },
                    'flag': {
                        'type': 'string',
                    },
                    'approved': {
                        'type': 'boolean',
                        'coerce': to_bool
                    }
                }
            }
        } 
    }
}
```

dkellner:
I've looked into this and can confirm your problem using Cerberus 0.9.2 (the bug is actually in the Cerberus validation library and not in Eve). It is already fixed in the current stable version (1.0.1), see https://github.com/nicolaiarocci/cerberus/issues/252. But unfortunately Eve does not support Cerberus 1.0 yet, see https://github.com/nicolaiarocci/eve/issues/776.

gwainer:
Thanks for the update @dkellner. Cerberus 1.0 is a big improvement indeed (I miss also the `excludes` rules). Now watching #776.

Issue #574, opened 2015-03-15
Integrate with swagger?


Jonovono:
Hi. I am surprised I have not found anyone asking this yet. 

Is there a way to get EVE working with the swagger library, or swagger-ui? (https://github.com/swagger-api/swagger-ui) It would be cool to have it work together so you could see a browsable api.

rs:
+1

nicolaiarocci:
Hello, it looks like a cool project of which, honestly, I didn't know until today :smile: 

We have something similar with the [Eve-Docs](https://github.com/charlesflynn/eve-docs) extension, although at first glance I'd say that swagger has a much broader scope/feature set.

MoOx:
I would love to be able to use swagger-ui with eve :heart:

harobed:
> I would love to be able to use swagger-ui with eve :heart:

+1

bcorso:
+1. One immediately useful feature swagger has over eve-docs is the ability to test the api request/response right from the doc (live demo: http://petstore.swagger.io/#/ )

dtheodor:
:+1:

toxsick:
I would love this :+1:

andy37380:
swagger-codegen is also cool :+1: 

[swagger-codegen](https://github.com/swagger-api/swagger-codegen)
> swagger-codegen contains a template-driven engine to generate client code in different languages by parsing your Swagger Resource Declaration

yxlwfds:
+1

nicolaiarocci:
I started some experimental work on a eve-swagger extension. See https://github.com/nicolaiarocci/eve-swagger.

nicolaiarocci:
Still a WIP, but usable already. Minimal output is validated by Swagger tooling:

![API docs in Swagger Editor](https://raw.githubusercontent.com/nicolaiarocci/eve-swagger/master/resources/swagger_editor.png)

Cheers

cbocta:
any news from that feature? :-)

nicolaiarocci:
@cbocta [eve-swagger](https://github.com/nicolaiarocci/eve-swagger) is in a "usable" state already. Wish I had the time to work some more on it, but the community can help there too :)

PR #702, opened 2015-09-08
Add 'SCHEMA_ENDPOINT' option

nckpark:
Add 'SCHEMA_ENDPOINT' option, allowing resource schema to be returned from an API endpoint. Defaults to None.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -183,6 +183,8 @@ Version 0.6
 - Fix: Geopoint coordinates do not accept integers. Closes #591 (Joakim
   Uddholm.)
 - Fix: OpLog enabled makes PUT return wrong Etag. Closes #590.
+- New: ``SCHEMA_ENDPOINT`` allows resource schema to be returned from an API
+  endpoint
 
 Stable
 ------
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -588,6 +588,9 @@ uppercase.
                                     and document changes are also logged to the
                                     :ref:`oplog`. Defaults to ``True``.
 
+``SCHEMA_ENDPOINT``                 Name of the :ref:`schema_endpoint`. Defaults
+                                    to ``None``.
+
 ``HEADER_TOTAL_COUNT``              Custom header containing total count of 
                                     items in response payloads for collection
                                     ``GET`` requests. This is handy for ``HEAD``
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -1875,6 +1875,19 @@ query the oplog for changes occured on that endpoint.
     in case you are wondering yes, the Eve oplog is blatantly inpsired by the
     awesome `Replica Set Oplog`_.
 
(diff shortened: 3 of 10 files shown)


nicolaiarocci:
Interesting. Care to elaborate on the use cases that might take advantage from this feature? Looks similar to what #574 would like to achieve, but with Swagger support.

nckpark:
I'm not familiar with Swagger, but at first glance - yeah, I think the inspiration is similar, and this feature could be evolved to offer Swagger spec data as a schema format option.

My immediate use case for this is to allow an API client library to fetch and parse schema to support automatically resolving data_relations and other schema constructs and construct empty documents, all without having to maintain duplicate field and type information in the library itself.

nckpark:
What are your thoughts on this? Any more questions?

nicolaiarocci:
I am a little reluctant to adding this. [Eve-docs](https://github.com/charlesflynn/eve-docs) already provides insights on schema definitions and, besides that, proper Swagger support would provide a better, standardised and widely adopted solution.

nicolaiarocci:
PS: look at the JSON-output section of the Eve-Docs readme.

nckpark:
Eve-docs does provide schema insight - more than I thought in fact - but it doesn't fully cover my use case because it doesn't provide any schema for read only resources. (Which a schema aware client needs to still allow users to interact with fetched data, even if they can't save it.)

As for Swagger, I don't see this as an only one or the other situation. Swagger support can be added in addition to a raw schema option. The benefits of offering both:
- Developers not building or using Swagger clients have a method to access schema data as needed, without the overhead of Swagger specification parsing and translation.
- Complex schema or validation rules which don't map well to the Swagger spec format can still be provided via the API
- Validation and any other schema processing code can be shared between client and server
- Returning raw schema is simple, and already done : )

All of that being said - it's your call. If you prefer not to include this in Eve, I understand, and it's simple enough for developers to implement at the application level.

nicolaiarocci:
Sounds good to me. I'll merge as soon as 0.6 is out of the way though.

nckpark:
Thanks! When is the 0.6 release planned?

nicolaiarocci:
The plan is Soon (next week, I hope). There's currently one issue with Cerberus and Python 2.6 holding it  down #701.

nckpark:
Sounds good. We currently use an internal fork of Eve to run stable development versions, but are hoping to move back to official releases as soon as we can. What do you think the timeframe is before this SCHEMA_ENDPOINT feature makes it into an official release?

nicolaiarocci:
Aww screw it I'm just gonna merge for 0.6 :smile:

nckpark:
Thanks!  : D

nicolaiarocci:
Merged in https://github.com/nicolaiarocci/eve/commit/21ac82b55e599d42c385f3933bbf3323ea37133b

PR #707, opened 2015-09-09
create_index add indexes to the versions collection

kalbasit:
`create_index` currently add indexes only for the resource but we also require indexes on the resource`_versions` which is created if versioning is enabled.

### DIFF
--- a/eve/io/mongo/mongo.py
+++ b/eve/io/mongo/mongo.py
@@ -928,22 +928,25 @@ def create_index(app, resource, name, list_of_keys, index_options):
         if any(auth):
             db.authenticate(username, password, source=auth_db_name)
 
-    coll = db[collection]
-
     kw = copy(index_options)
     kw['name'] = name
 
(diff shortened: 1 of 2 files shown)


nicolaiarocci:
Hey thanks for this, really needed. 

Would you refactor it a little bit, for increased DRY. [These lines](https://github.com/kalbasit/eve/blob/create_indexes_for_version_collection/eve/io/mongo/mongo.py#L931-L944) could go into a function or closure, so they can be invoked twice instead of repeating them for both standard and versioned collection.

pfreixes:
Now these lines [1] need to become a closure.
@nicolaiarocci +1 

[1] https://github.com/kalbasit/eve/blob/create_indexes_for_version_collection/eve/io/mongo/mongo.py#L931-L944

kalbasit:
PTAL

nicolaiarocci:
Excellent. Thank you.

Issue #682, opened 2015-08-11
is argument "updates" meant to be "updated" for on_updated call


guonsoft:
In patch.py, "updates" is used when calling on_updated callbacks.  Is it meant to be "updated" instead?  "update" has done its work and won't be used anymore in building the response.

            getattr(app, "on_updated")(resource, updates, original)
            getattr(app, "on_updated_%s" % resource)(updates, original)

nicolaiarocci:
Well I guess it's all about semantics: `updates` is meant to mean "updates requested" or "updates performed", which is appropriate I believe?

guonsoft:
Sorry I might not have described it clearly.  I think it is a typo/bug otherwise the on_updated* hooks won't have any effect to the final response.  By the time the code gets to this point, it has both documents "update" and "updated".  "updated" contains the result and is the document feeding into build_response_document() to build full response.

nicolaiarocci:
Oh I see what you mean now. You would rather have the updated document back in your callback function, rather than the updates sent with the request.

sphet:
I know this is closed, and I hate to necro it, but I don't understand how I am supposed to get the complete updated item record in the callback. It seems like this patch update passing back the updated item. What is the correct way to get the updated item - is it to take the original and update it with updates myself in the callaback?

PR #709, opened 2015-09-10
Fix regression for on_updated

amagdas:


nicolaiarocci:
This was actually intended. See https://github.com/nicolaiarocci/eve/blob/develop/CHANGES#L23-L24.

amagdas:
I know but it looks like a regression... What it the actual use case for it? If you get only the payload, you can easily obtain the 'updated' as the 'original' is passed as well.

amagdas:
The build is passing, the rate limit heisenbug is doing it's work.

petrjasek:
it allows you to modify the response, we use that somewhere (i think with dictionaries)

nicolaiarocci:
Also see discussion at #682

amagdas:
My fix it's actually fixing #682 , we had this fix in our fork for quite some time, didn't manage to create a proper PR for it though :(

amagdas:
Basically this PR is consistent, meaning you can still easily know what the payload is when calling 'on_updated' and changes to 'updates' will be reflected in the response.

nicolaiarocci:
oh right, I didn't notice the final update call.

amagdas:
So is this a proper fix?

nicolaiarocci:
Yes. Merged with https://github.com/nicolaiarocci/eve/commit/43dfe08d6008114a19628ef7d6f18652ba401afb

Thank you!

PR #714, opened 2015-09-17
Feature: add the ability to customize the returned HATEOAS links

amagdas:
We use it for returning in  a single call entities from multiple resources


nicolaiarocci:
thanks!

PR #716, opened 2015-09-23
Fix pymongo regression

amagdas:


nicolaiarocci:
Thanks. I updated it to 2.8.1 which is the latest release Flask-PyMongo is compatible with I think.

PR #718, opened 2015-09-24
Add the ability to embed ObjectId directly

goneri:
So far it was only possible to embed list, with this patch it's now
possible to embed directly a given resource.
The patch also breaks the maximum number of level of embedded resources.


nicolaiarocci:
Super! This deserves a last minute inclusion in upcoming v0.6.

Thank you.

PR #724, opened 2015-09-30
move _client_projection() from mongo to DataLayer

goneri:
_client_projection() is generic. There is no reason to keep it in
the mongodb drive. By moving it on the DataLayer class we will be
able to drop some (oudated) duplicated code from Eve-SQLAlchemy:

https://github.com/RedTurtle/eve-sqlalchemy/blob/0.4.0a2/eve_sqlalchemy/__init__.py#L325-L340

### DIFF
--- a/eve/io/base.py
+++ b/eve/io/base.py
@@ -15,7 +15,7 @@ from copy import copy
 from flask import request, abort
 from eve.utils import date_to_str
 from eve.auth import auth_field_and_value
-from eve.utils import config, auto_fields
+from eve.utils import config, auto_fields, debug_error_message
 
 
 class BaseJSONEncoder(json.JSONEncoder):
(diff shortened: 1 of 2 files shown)


eve/methods/common.py
```diff
@@ -1047,3 +1047,23 @@ def oplog_push(resource, document, op, id=None):
 
     if entries:
         app.data.insert(config.OPLOG_NAME, entries)
+
+    def _client_projection(self, req):
```
nicolaiarocci: isn't this a duplicate of what has now been moved to `eve/io/base.py`?

goneri: hum, what part?

My plan is to adjust this function to be able to deal with different level of projection, like:
```projection={"my_embedded_resource.key":1}.```
I may be wrong, but I don't think this is something that work for the moment.

nicolaiarocci: what I mean is that your PR has two `_client_projection` methods. One is in `eve.io.base` and another in `eve.methods.common`, which appears to be not only superfluous but also orphaned. I think that the latter should be removed (probably a leftover from experimenting with the code move)

goneri:
outch. fixed!

nicolaiarocci:
Merged with https://github.com/nicolaiarocci/eve/commit/333049e9d71ab6e88ae8a68f2ad5a3e2adb13ed0

Issue #728, opened 2015-10-02
Validating the where clause in Filtering


stratosgear:
I have a use case where I would like the ?where= statement on a GET request be validated against the Resource schema and, obviously, fail if the validation does not pass.

I've posted a StackExchange question:

http://stackoverflow.com/questions/32611654/in-python-eve-how-do-i-validate-the-where-clause

but I didn't receive any valuable feedback, except for the fact that I can **manually** implement it through a hook event.

I find such functionality essential to Eve (well, at least if you ask **me**).  

First off, is something similar offered already, that I have overlooked?

And second, is it something that is worth creating a PR for...? Maybe by providing a setting if you want the where to be validated...?  I'd hate to cook something and then never have it merged. :)

nicolaiarocci:
> I find such functionality essential to Eve (well, at least if you ask me).

I guess it could be a nice feature to have.

> First off, is something similar offered already, that I have overlooked?

No.

> And second, is it something that is worth creating a PR for...? Maybe by providing a setting if you want the where to be validated...? I'd hate to cook something and then never have it merged. :)

Yes, thanks, let's see what you come up with. I'm more than willing to review and assist. I'm not sure a setting would be necessary, but certainly it cannot be a bad thing and it would have very little impact so go for it.

PR #729, opened 2015-10-04
Validate get filters against resource schema

stratosgear:
This should take car of #728.

I included a setting called VALIDATE_FILTER, since the existing code made the assumption that the filter values can be anything, so if the default was to check the validity of the filtering, then a ton of test cases were failing.

Also added some extra test cases, with nested resource attributes and ($and, $or, $nor) operators.

### DIFF
--- a/eve/tests/methods/get.py
+++ b/eve/tests/methods/get.py
@@ -899,6 +899,56 @@ class TestGet(TestBase):
         response, status = self.get(self.known_resource, where)
         self.assert400(status)
 
+    def test_get_nested_filter_operators_unvalidated(self):
+        """ test that nested filter operators are working correctly.
+        """
+        where = ''.join(('?where={"$and":[{"$or":[{"fldA":"valA"},',
+                    '{"fldB":"valB"}]},{"fld2":"val2"}]}'))
+        response, status = self.get(self.known_resource, where)
+        self.assert200(status)
+
+    def test_get_nested_filter_operators_validated(self):
+        """ test that nested filter operators are working correctly.
+        """
+        self.app.config['VALIDATE_FILTERS'] = True
+
+        where = ''.join(('?where={"$and":[{"$or":[{"fldA":"valA"},',
+                    '{"fldB":"valB"}]},{"fld2":"val2"}]}'))
+        response, status = self.get(self.known_resource, where)
+        self.assert400(status)
+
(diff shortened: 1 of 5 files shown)


nicolaiarocci:
Thanks looks good except a couple flake8 issues.

stratosgear:
You know, I have done 5-6 updates trying to fix this flake8 issue with wrapped long lines, but I can't get it to NOT complain.  I cannot find any exact docs, on how to fix the issue too.  So far I have figured it is the indentation level that is off, but after trying some sane values, it still complains.  Sorry!

eve/tests/methods/get.py
```diff
@@ -899,6 +899,56 @@ def test_get_allowed_filters_operators(self):
         response, status = self.get(self.known_resource, where)
         self.assert400(status)
 
+    def test_get_nested_filter_operators_unvalidated(self):
+        """ test that nested filter operators are working correctly.
+        """
+        where = ''.join(('?where={"$and":[{"$or":[{"fldA":"valA"},',
```
julianhille: i really don't like how this where clause gets created.

stratosgear: I agree! I also don't like it at all.

Any suggestions how this string:

    ?where={"$and":[{"$or":[{"fldA":"valA"},{"fldB":"valB"}]},{"fld2":"val2"}]}

or this (even prettier):

    ?where={
      "$and": [
        {
          "$or": [
            {
              "fldA": "valA"
            },
            {
              "fldB": "valB"
            }
          ]
        },
        {
          "fld2": "val2"
        }
      ]
    }

can be typed as a python statement that passes flake8 line length conventions and doesn't use a ton of concatenations?

I tried multiple combinations, that none were prettier and decided to move along into more interesting things than keep playing with flake8... :)

julianhille: i would do something like:

```
where = {
    "$and": [
        {"$or": [
            {"fldA": "valA"},
            {"fldB": "valB"}
        ]},
        {"fld2": "val2"}
    ]}
self.get(self.known_resource, '?where={}'.format(json.dumps(where)))
```

nicolaiarocci:
Merged with https://github.com/nicolaiarocci/eve/commit/cd69e15116a441ee8d32a08cae99c37383f2df1c

Thank you!

PR #732, opened 2015-10-09
Build specific versions without deepcopying root document. 

nckpark:
Improves versioned document GET performance at load and for large documents.

Deep copies are very expensive, and not actually required when synthesizing a document version.


nicolaiarocci:
Thanks.

Issue #738, opened 2015-10-16
Dotted field expansion fails on certain nested inputs


ultimateboy:
There are two distinct recursion/nesting issues with the [normalized_dotted_fields()](https://github.com/nicolaiarocci/eve/blob/develop/eve/methods/common.py#L400) function.

First, once it finds a dotted-field, it does not continue down the tree looking for other dotted-fields. `{'a.b': {'c.d': 42}}` is normalized to `{'a': {'b': {'c.d': 1}}}`, missing the nesting of "c.d".

Second, the function does not correctly handle lists which contain dotted-fields. `{'a': [{'b.c': 42}]}` is unaltered when being normalized.

ultimateboy:
Pull request: https://github.com/nicolaiarocci/eve/pull/739

ultimateboy:
I'm at a loss as to why the test is failing. Would love some advice.

PR #740, opened 2015-10-16
Pass transparent_schema_rules option to Cerberus Validator

kynan:
This allows enabling transparent_schema_rules by setting either the global option TRANSPARENT_SCHEMA_RULES or the resource level option transparent_schema_rules.

Note that this option *must* be passed to the superclass constructor and cannot be set afterwards since validation of the schema happens on construction. This requires some minor refactoring of the eve.io.mongo.Validation constructor.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -11,6 +11,8 @@ Version 0.6.1
 
 Not released.
 
+- New: ``TRANSPARENT_SCHEMA_RULES`` enables/disables schema validation
+  globally and ``transparent_schema_rules`` per resource (Florian Rathgeber)
 - New: ``ALLOW_OVERRIDE_HTTP_METHOD`` enables/disables support for overriding
   request methods with ``X-HTTP-Method-Override`` headers (Julian Hille).
 - Fix: dependendencies on sub-document fields always return 422. Closes #706.
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -328,6 +328,9 @@ uppercase.
                                     :ref:`unknown` for more information.
                                     Defaults to ``False``.
 
+``TRANSPARENT_SCHEMA_RULES``        When ``True``, this option globally disables
+                                    :ref:`schema_validation` for any API endpoint.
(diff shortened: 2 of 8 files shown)


nicolaiarocci:
Thanks for this. I'm fine to merging but I suspect that in its current state it would break backward compatibility for some custom validators. Moving the constructors' `allow_transparent_rules` argument to the last position would be ideal. What do you think?

Also, default settings test needs an update and a specific test for transparent rules would be welcome.

kynan:
Tests are coming as well as documentation updates. I'll swap the argument order.

kynan:
@nicolaiarocci Added tests and documentation and moved `transparent_schema_rules` last.

nicolaiarocci:
:+1:

kynan:
Minor fixup (a ref in the docs was wrong). Good to merge now I think.

kynan:
I think the test failure is unrelated to my changes (I also can't reproduce it on my machine).

nicolaiarocci:
Yes i think that #738 introduced an intermittently failing test, which is quite annoying. Hope to have time to look at it, or maybe @ultimateboy can do that.

ultimateboy:
The intermittent failing of tests is incredibly odd. I've seen successful test runs with the changes from #738 included. But for some reason, the test added in that issue seems to fail randomly. I have no idea what's going on.

nicolaiarocci:
I've also seen it fail on my local tox environment. it didn't fail out of tox though. Weird. Won't have time to look at it until next week I'm afraid.

kynan:
@nicolaiarocci Anything holding this back still?

nicolaiarocci:
Nope it's fine and I've also fixed that broken test introduced with #738. Merge coming up.

nicolaiarocci:
Merged with https://github.com/nicolaiarocci/eve/commit/2209148f5bcaede3f2bfc4b76cc7954ccc35ad65

PR #748, opened 2015-10-23
Use python 2 for flake8-tests

lynxed:
There's an assignment of 'basestring' for to compatibility of both
python 2 and python 3. This causes flake8 to fail when executed with python 3
due to an unknown name.
This commit configures tox to run flake8 with python 2 even on systems
were python 3 is the default, to avoid problems here.


nicolaiarocci:
Excellent, thanks!

Issue #749, opened 2015-10-25
[pymongo3] Bump pymongo and Flask-PyMongo versions


g----:
Flask-Pymongo 0.4 was released few days ago. Time for updating requirements in pymongo3 branch.

nicolaiarocci:
the `pymongo3` branch has been updated. However, there is a [bug in PyMongo 3.0+](https://jira.mongodb.org/browse/PYTHON-983) which prevents its adoption on Python 2. 

The bug is fixed in PyMongo 3.1rc0, so when 3.1 is released we will finally be able to move to PyMongo 3.0 (for the long story see https://github.com/nicolaiarocci/eve/commit/a20873528d2f3b39d5c2e675bc99e640c886ac0a)

nicolaiarocci:
Support for PyMongo 3 has just been merged to the `develop` branch.

PR #750, opened 2015-10-25
#749 bump versions of pymongo and flask-pymongo to enable (py)mongo 3 support

g----:
Bumped requirements version for pymongo3 support.

Are we ready to merge pymongo3 to master?

### DIFF
--- a/eve/io/mongo/mongo.py
+++ b/eve/io/mongo/mongo.py
@@ -18,6 +18,7 @@ import simplejson as json
 from bson import ObjectId
 from flask import abort, request
 from flask.ext.pymongo import PyMongo
+from pymongo import WriteConcern
 from werkzeug.exceptions import HTTPException
 
 from eve.auth import resource_auth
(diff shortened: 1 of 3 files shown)


nicolaiarocci:
Ah nice it was about time they released it :) 

Right now however there are performance issues with the test suite. I hope to have time to look at them soon, and I'm  not sure  the issue is on eve side actually, something fishy might be going on with latest Flask-PyMongo release. I updated the pymongo3 branch if you or somebody else wants to investigate it further.

nicolaiarocci:
Ok I found the culprit. There are performance issues with PyMongo 3.0.x and Python 2.x. They have been fixed in PyMongo 3.1 which is not available yet so we'll have to wait until that one is released. For details see https://github.com/nicolaiarocci/eve/commit/b25bdf31b57bf5360fec543b9ccb4a20534750b4

g----:
Ok, let's wait for pymongo 3.1 then.

Thanks for investigating!

nicolaiarocci:
Support for PyMongo 3 is now merged to Eve-dev.

PR #756, opened 2015-11-05
Allow usage of mongo prefix even from flask route.

xgvargas:
This fix allows to select a mongo prefix even if you are not using an Eve resource. 

So, even from an `@app.route` is now possible to use: 

````python
from flask import g

g.mongo_prefix = 'MONGO2'
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -41,6 +41,7 @@ Patches and Contributions
 - Gino Zhang
 - Gonéri Le Bouder
 - Grisha K.
+- Gustavo Vargas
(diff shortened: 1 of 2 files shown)


nicolaiarocci:
Looks good. Would you update the docstring to illustrate how one would use `g` from a custom function?

nicolaiarocci:
Merged with https://github.com/nicolaiarocci/eve/commit/2d99faf80917456d6217d65e786630395356bd7d

Thanks

PR #762, opened 2015-11-18
Allow other parameters to be part of urls

mugurrus:
Parameters not recognised by Eve are stripped from HATEOAS URLs.
This change adds those parameters to HATEOAS URLs.

E.g.: I want to implement a different way of filtering content,
instead of using where:
?start_date=2014-01-25&search_for=foo

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -81,6 +81,7 @@ Patches and Contributions
 - Mayur Dhamanwala
 - Mikael Berg
 - mmizotin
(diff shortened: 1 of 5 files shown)


nicolaiarocci:
Looks good, but it needs test coverage.

nicolaiarocci:
Excellent, thank you.

mugurrus:
Thank you!

PR #764, opened 2015-11-19
Option to write full object to the oplog on POST

mmiz:
I use oplog to track object change history. Currently, the 'c' field is not written on document creation via POST, thus it is hard to view the initial document state.

This pull request makes methods which write changes to oplog configurable with option OPLOG_CHANGE_METHODS, with the default value ['DELETE', 'PATCH', 'PUT'] corresponding to the current behavior.

I'm not sure about the naming though.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -597,6 +597,10 @@ uppercase.
                                     Defaults to ``['DELETE', 'POST', 'PATCH',
                                     'PUT']``.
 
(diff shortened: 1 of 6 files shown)


nicolaiarocci:
Merged with https://github.com/nicolaiarocci/eve/commit/231d087d96a1ea861caeaafee6bfaab34d78237c

Thanks!

PR #768, opened 2015-11-24
Validator is not set when skip_validation is true

stratosgear:
Fixes issue introduced with 0b2ae5a928c18f53c598ae2ba4536df6f9e175f4 in which case when the ```patch_internal``` method was called with ```skip_validation=True``` the ```validator``` on line 139 will *not* be instantiated, but it will try to be accessed on line 163 causing a:

      File "/usr/local/lib/python2.7/dist-packages/eve/methods/patch.py", line 163, in patch_internal
        updates = validator.document
    UnboundLocalError: local variable 'validator' referenced before assignment
    ERROR:eve:local variable 'validator' referenced before assignment

### DIFF
--- a/eve/methods/patch.py
+++ b/eve/methods/patch.py
@@ -135,8 +135,7 @@ def patch_internal(resource, payload=None, concurrency_check=False,
 
     resource_def = app.config['DOMAIN'][resource]
     schema = resource_def['schema']
-    if not skip_validation:
-        validator = app.validator(schema, resource)
+    validator = app.validator(schema, resource)
 
     object_id = original[resource_def['id_field']]
     last_modified = None
--- a/README.rst
+++ b/README.rst
@@ -8,7 +8,7 @@ allows to effortlessly build and deploy highly customizable, fully featured
 RESTful Web Services.
 
 Eve is powered by Flask, Redis, Cerberus, Events and offers support for both
-MongoDB and SQL backends.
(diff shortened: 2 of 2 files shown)


eve/methods/patch.py
```diff
@@ -135,8 +135,7 @@ def patch_internal(resource, payload=None, concurrency_check=False,
 
     resource_def = app.config['DOMAIN'][resource]
     schema = resource_def['schema']
-    if not skip_validation:
-        validator = app.validator(schema, resource)
+    validator = app.validator(schema, resource)
```
nicolaiarocci: Good catch, but won't `validator.document` be `None` on line 162, when `skip_validation = True`? Can't try right now but if I am correct then a better change would be to move 162 in the `else` block starting at line 157.

stratosgear: It's either/or! 

My intend was to keep the ```validator``` instantiated anyways (regardless of skip_validation or not) so if it needs to be used *anywhere* in the method (even in multiple places) it would be available.

Unless instantiating it incurs a huge cost (don't think so) I feel it is cleaner to have it available throughout the method.

You have the final say, off course!

nicolaiarocci:
Marged with https://github.com/nicolaiarocci/eve/commit/0db773175d7db9c517883b3f781da94ee17842d6

Thanks.

PR #773, opened 2015-11-27
Skip any null value from attempting serialization (Fixes #772)

### LINKED ISSUE #772: Serialization procedure does not support nullable lists
If a nullable list schema contains a serializable item, sending a null value for it raises an exception and serialization is ended prematurely.

arnauorriols:
On a side note, given the strong contract that this PR enforces, the null guards in the serializers could (and should, IMHO) be removed.


nicolaiarocci:
Thanks!

Feel free to remove the guards. Let's see if the castle of cards holds strong or not :smile:

PR #774, opened 2015-11-27
Add missing serializer for fields of type number.

arnauorriols:
The serializers of int, float and number are required when using
Eve through x-www-urlencode requests. Added some test cases that
covers this dependency.


nicolaiarocci:
Thanks. Merged with 347cee6c588185119a2f3172123a1848faf1af24

PR #777, opened 2015-12-04
Return 428 instead of 403 when If-Match header is missing (closes #775)

### LINKED ISSUE #775: Consider 428 when If-Match header is missing?
Hi there.

I just had a hard time debugging an issue of a client, because the 403 response code is just too broad (IMO) and neither my client could realize what was she doing wrong by herself, nor I remembered to check the possibility that 403 could mean the If-Match header missing (it is worth noting that our product does an extensive use of the 403 response to denote permissions failure).

As per the [RFC6585](http://tools.ietf.org/html/rfc6585#section-3), there is an status code dedicated to this particular issue:

> 3.  428 Precondition Required
>
> The 428 status code indicates that the origin server requires the
> request to be conditional.
>
>  Its typical use is to avoid the "lost update" problem, where a client
>  GETs a resource's state, modifies it, and PUTs it back to the server,
>  when meanwhile a third party has modified the state on the server,
>  leading to a conflict.  By requiring requests to be conditional, the
>  server can assure that clients are working with the correct copies.
>
>  Responses using this status code SHOULD explain how to resubmit the
>  request successfully.  For example:
>
>  HTTP/1.1 428 Precondition Required
>  Content-Type: text/html
>  \<html\>
>     \<head\>
>        \<title\>Precondition Required\</title\>
>     \</head\>
>     \<body\>
>        \<h1\>Precondition Required\</h1\>
>        \<p\>This request is required to be conditional;
>        try using "If-Match".\</p\>
>     \</body\>
>  \</html\>

What do you think about this?

arnauorriols:


nicolaiarocci:
Merged with https://github.com/nicolaiarocci/eve/commit/1ac43cd754f70a9725094616bf3b518e68e04fce

Thanks!

PR #799, opened 2016-01-06
Fixes #797 if schema is None in a validator

### LINKED ISSUE #797: validator schema not needed but fails
The validator class in eve.io.mongo.validation has an inconsistency in its requirements:
```
 def __init__(self, schema=None, resource=None, allow_unknown=False,
                 transparent_schema_rules=False):

```

States that schema defaults to None and is not required.
```
self = <play.api.application.Validator object at 0x1081a7eb8>, schema = None

    def _remove_unique_rules_on_fields_with_unique_index(self, schema):
        # TODO: Actually do what the function name suggests. This version just
        # removes the unique constraint on _id. We could use the information
        # available by app.data.driver.db[datasource].index_information() to
        # remove all unnecessary unique constraints.
        result = copy.deepcopy(schema)
>       if '_id' in result and 'unique' in result['_id']:
E       TypeError: argument of type 'NoneType' is not iterable
```

So two ways of fixing that from here imho
1. make schema a  required parameter
2. ignore schema if None in _remove_unique_rules_on_fields_with_unique_index

julianhille:


nicolaiarocci:
See my #797 note.

eve/io/mongo/validation.py
```diff
@@ -71,10 +71,10 @@ def _remove_unique_rules_on_fields_with_unique_index(self, schema):
         # removes the unique constraint on _id. We could use the information
         # available by app.data.driver.db[datasource].index_information() to
         # remove all unnecessary unique constraints.
-        result = copy.deepcopy(schema)
```
nam4dev: result = copy.deepcopy(schema or {}) is enough :)

julianhille: @nam4dev i know but two things kept me from doing so:
1. unneeded function call to copy an empty dict
(but further more)
2. unneeded deepcopy of schema even if schema['_id']['unique'] does not exists.
This could be very expensive operation if schema is HUGE and that nested key does not exist.

nam4dev: @julianhille Right :) That's a good point!

What about that?
* has_unique_constraints = (schema or {}).get('_id', {}).get('unique')

Another point, why are you reassigning the schema *deepcopy*  to the schema itself ?
**Instead of**
* schema = copy.deepcopy(schema)

**Something more like**
* schema_copy_or_whatever_make_sense = copy.deepcopy(schema)

Don't you think or am I missing something?

nicolaiarocci:
Addressed with 8e2e17d3e02c3c0a95a5f417fcf6cf5a36cd8dc3

PR #801, opened 2016-01-11
allow the options method on the schema endpoint

valerierose:
Fixes issue described here: http://stackoverflow.com/questions/34400555/how-to-call-a-get-on-schema-endpoint-with-cors-in-eve


nicolaiarocci:
Thanks for this! Would be great if you could add a test so we don't have a regression in the future.

nicolaiarocci:
Merged with https://github.com/nicolaiarocci/eve/commit/4fdb678de55b19474ec95950bcb49b1cf51714dc. Thanks!

PR #804, opened 2016-01-13
Don't attempt to parse number values as strings when they are numerical.

nckpark:
Also standardizes value serialization exception handling across the serialize function.

An [earlier commit](https://github.com/nicolaiarocci/eve/commit/347cee6c588185119a2f3172123a1848faf1af24) added a new serializer for values of field type `number` which parsed the values as JSON strings in order to support `x-www-urlencode` requests. However, those values are not strings when submitted in more typical `application/json` requests. The parsing should only happen when the given value is a string.

Attempting to parse actual numbers throws an exception which was sometimes caught by a try/except block around the serializer call, but that exception handling only wrapped one possible serializer call. When submitting more complex schema that hit other serializer calls, a list of numbers for example, the exception would abort document serialization with the error squashed at the `parse` level. This would leave the document partially un-serialized, most commonly resulting in un-cast ObjectId's causing validation to fail. In order to avoid legitimate value serialization errors from causing similar issues I've updated the serializer calls to use a method which always applies exception handling.

### DIFF
--- a/eve/methods/common.py
+++ b/eve/methods/common.py
@@ -369,15 +369,14 @@ def serialize(document, resource=None, schema=None, fields=None):
                                     serialize(sublist[i],
                                               schema=sublist_schema['schema'])
                                 elif item_type in app.data.serializers:
-                                        sublist[i] = \
-                                            app.data.serializers[item_type](v)
+                                    sublist[i] = serialize_value(item_type, v)
                     else:
                         # a list of one type, arbitrary length
                         field_type = field_schema.get('type')
                         if field_type in app.data.serializers:
                             for i, v in enumerate(document[field]):
                                 document[field][i] = \
-                                    app.data.serializers[field_type](v)
+                                    serialize_value(field_type, v)
                 elif 'items' in field_schema:
                     # a list of multiple types, fixed length
                     for i, (s, v) in enumerate(zip(field_schema['items'],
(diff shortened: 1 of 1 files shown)


eve/io/mongo/mongo.py
```diff
@@ -66,7 +66,8 @@ class Mongo(DataLayer):
         'datetime': str_to_date,
         'integer': lambda value: int(value) if value is not None else None,
         'float': lambda value: float(value) if value is not None else None,
-        'number': lambda val: json.loads(val) if val is not None else None
+        'number': lambda val: json.loads(val) if isinstance(val, basestring)
```
nicolaiarocci: This will fail under Python 3.x. Try importing and using `base_str`  from `utils` ([link](https://github.com/nicolaiarocci/eve/blob/develop/eve/utils.py#L436)).

nckpark:
Update to use str_type in response to the Python 3.x basestring problem. I hadn't run the full set of environments before submitting this - thanks for the catch.

Looking at it again, I realized that because we're now always wrapping the serialization in exception handling, we could not do any type checking and just rely on duck typing here. Might be a tiny bit more efficient? Interested in your opinion.

nicolaiarocci:
hmm might make sense. Should also simply the code a little bit. Worth a try.

nckpark:
OK - reverted back to always attempting the JSON parsing of number fields, with duck typing through consistent exception handling allowing both strings and numerical values to be serialized correctly.

nicolaiarocci:
Thanks!

Issue #806, opened 2016-01-14
Resource field validation fails while submitting resources with "multipart/form-data"


stratosgear:
I had a Resource with a certain amount of fields defined as booleans, numbers and string lists.  Field validation was working correctly and EVE was rejecting resources with incorrect field types.

Later on I found out that I needed to have a binary file attached to the resource.  Reading how Eve handles media files (media can be any binary file) I was forced to:
* change my POST operation from 'application/json' (just submitting a big JSON object in the body of the POST)
* to a "multipart/form-data" in order to accommodate the file upload

What I did was to break the big JSON object to separate form data fields and try to submit them like JSON encoded strings along with the uploaded file.

So a fields like ```tags``` defined as:

    'tags': {
        'type': 'list',
        'schema': {'type': 'string'}
    }

would be submitted as:

    tags = '["tagA", "tagB"]'

Unfortunately, it seems that EVE treats **all** incoming form data as ```strings``` therefore when it is time to validate them it would complain stating that the field had to follow the proper schema and be submitted as a list.  In the same manner, **ALL** fields that were not defined as ```strings``` would also fail validation.

I was forced to make some changes to EVE in order to accommodate my use case.

I will be shortly submitting a PR to address this issue.

PR #807, opened 2016-01-14
FIX: Validate resource fields submitted as multipart/form-data

stratosgear:
Scratching my own #806 itch

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -22,6 +22,10 @@ Version 0.7
 - Fix: ETag request headers which conform to RFC 7232/2.3 (double quoted value)
   are now properly processed. Addresses #794.
 
+- Fix: Incoming resource parameters that were submitted through
+  multipart/form-data are correctly validated, instead of assumed that they
+  are all of type "string". Addresses #806 (Stratos Gerakakis)
+
 
 Version 0.6.2
 ~~~~~~~~~~~~~
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -585,6 +585,19 @@ uppercase.
                                     dedicated media endpoints. Defaults to
                                     ``regex("[a-f0-9]{24}")``.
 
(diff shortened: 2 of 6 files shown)


nicolaiarocci:
Cool stuff. 

1. Tests are failing
2. Wouldn't it be nice if this feature would kick in automatically when an endpoint has `media` type files in its schema? Just an idea, have not looked into it at all, but at first sight I don't see why one would want to be capable of only parsing string types.

stratosgear:
I know about the failing tests. I'm having troubles bringing a 2.6 python env up to test why they fail just on 2.6, but pass on all others.

Unfortunately, I do not know how to make the process automatic, as you suggest.  I also agree that it doesn't make sense to have a resource full of strings.  You would still need to validate your resource even when you upload media.

The problem is with parsing the incoming form fields. As long as the incoming fields are properly formatted json structures everything is OK. Except if one wants to pass an actual string, in which case they **will have to enclose it in double quotes**, something that is not so obvious. The way the functionality is implemented is to parse all incoming form fields as JSON strings, a process where any random string (unless doubly quoted) fails. 

Not notifying the user and turning this feature on, will break backwards compatibility because no one encloses strings in double quotes by default

Kinda hard to explain....  Maybe someone else, smarter than me can step up, but I do not see any "transparent" way of handling it.  That's why I tried to implement it as an optional feature with a warning on how to properly deal with strings.

Do you have an issue label, labeled "Help Required" or something, so maybe more people might take a look?

nicolaiarocci:
Howdy. I fixed the Py26 issue, reabased and merged with https://github.com/nicolaiarocci/eve/commit/c8a3fdfb39e4ab41fc4a055cce21fcb5f686085b

Thanks!

nicolaiarocci:
PS: please double-check that I got the string syntax right while editing your documentation changes (http://python-eve.org/features.html#file-storage). If  not, please PR a fix.

PR #812, opened 2016-01-17
fix PUT validator issue with skip_validation is true, and add a test

wguan2:
Apply the same fixes that are already available in PATCH method.  Also add a test for PUT skip_validation=true


nicolaiarocci:
Merged with https://github.com/nicolaiarocci/eve/commit/120082f9733e0c36507540a7ac085127a54326de

Thanks!

Issue #776, opened 2015-12-02
Cerberus 1.0 Support


nckpark:
This issue is a bit premature as Cerberus 0.10 has yet to be released, but I'm curious what the plan is for updating Eve to use it? Is there any timeline in mind? I'm eager to see the impact on validation performance - it seems like there could be some significant improvements in the new code.

nicolaiarocci:
I plan on supporting Cerberus 0.10 because it is better on all accounts. I have been hesitant however since it breaks backward compatibility in a big way so it is going to require a bit of work and, which is worse, would probably break most custom validators making the upgrade to next Eve release a pain point. 

I have not had the time to experiment with the update though, so please feel free to play with it and submit a PR if you feel like doing it.

jim-bo:
I would like to second the request for 1.0 support. I think the advances in Cerberus 1.x would be great to have in Python-Eve. In my case the recursive schema definitions is needed.

dkellner:
I've started playing with this. It will probably take some time, though. I am planning to post an update on this within one or two weeks.

nicolaiarocci:
@dkellner excellent!

dkellner:
See https://github.com/nicolaiarocci/eve/pull/917

nicolaiarocci:
See #1001

PR #815, opened 2016-02-01
Allow callables for 'default' in schema

dkellner:
Allows the use of calculated default values via callables, e.g. lambda functions. They can even depend on one another and use all other values of the document, while circular dependencies are properly detected: a ```RuntimeError``` is raised in this case.

A small example how to use this:
```python
def one_hour_after_value_date(document):
    from datetime import timedelta
    return document['value_date'] + timedelta(hours=1)

RESOURCE_METHODS = ['GET', 'POST']

DOMAIN = {
    'transactions': {
        'schema': {
            'amount': {
                'type': 'integer',
                'required': True
            },
            'expiration_date': {
                'type': 'datetime',
                'default': one_hour_after_value_date
            },
            'value_date': {
                'type': 'datetime',
                'default': lambda doc: doc['_created']
            },
        }
    }
}
```

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1178,7 +1178,13 @@ defining the field validation rules. Allowed validation rules are:
 
 ``default``                     The default value for the field. When serving
                                 POST and PUT requests, missing fields will be
-                                assigned the configured default values.
+                                assigned the configured default values. If the
+                                default value is a callable, it is evaluated
+                                before assignment.
+
+                                Callables can even depend on one another, but
+                                a ``RuntimeError`` will be raised if there is a
+                                circular dependency.
 
                                 It works also for types ``dict`` and ``list``.
                                 The latter is restricted and works only for
--- a/eve/defaults.py
+++ b/eve/defaults.py
@@ -94,26 +94,51 @@ def resolve_default_values(document, defaults):
     .. versionadded:: 0.2
     """
     todo = [(defaults, document)]
+    circular_dependency_checker = CircularDependencyChecker()
     while len(todo) > 0:
-        defaults, document = todo.pop()
+        circular_dependency_checker.register_todo_list(todo)
+        defaults, document_part = todo.pop(0)
         if isinstance(defaults, list) and len(defaults):
-            todo.extend((defaults[0], item) for item in document)
+            todo.extend((defaults[0], item) for item in document_part)
             continue
         for name, value in defaults.items():
+            if callable(value):
+                try:
+                    value = value(document)
+                except KeyError:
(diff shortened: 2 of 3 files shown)


nicolaiarocci:
This is a great contribution. However, Cerberus 1.0 will bring new normalisation rules, including a [brand new `default` rule](https://cerberus.readthedocs.org/en/latest/normalization-rules.html#default-values). My current plan would be to replace Eve's own `default` implementation with Cerberus' built-in. This of course can only happen once Cerberus 1.0 is released. I'm hoping to support new Cerberus with Eve 0.7. 

If I recall correctly, right now Cerberus's `default` rule does not support callables, so it would be more appropriate to add support for them in there. There are some more issues however. Validation (and normalisation with v1.0) happens before document meta fields are initialised, so your code snippet would still not work..

Thoughts?

dkellner:
Hm, too bad :). I really like this feature, in fact, we're using this in production right now. But I think you're right in moving this logic to Cerberus. I will have a look into it in the next days. Maybe I can add callable support based on my changes here to Cerberus.

Do you mean the snippet below by initialization of document meta fields? If yes, we could try to set the future callable-supporting Cerberus-default-rule to a `current_utc_datetime` function and let Cerberus do the assignment. What do you think?

```python
# Populate meta and default fields
document[config.LAST_UPDATED] = \
    document[config.DATE_CREATED] = date_utc
```
(eve/methods/post.py, L197)
https://github.com/nicolaiarocci/eve/blob/2d9a69a8e32c566ad9e7a97d99686f7ab84b979c/eve/methods/post.py#L197

dkellner:
I did a similar implementation for Cerberus: https://github.com/nicolaiarocci/cerberus/pull/200.

If you agree, I would change this one to behave exactly the same as the Cerberus one. I.e. passing only the current (sub)document to the callable. This way we could manage backward compatibility if you decide to merge this before Cerberus 1.0 is released.

nicolaiarocci:
I'd rather hold onto this one until we merge new Cerberus. Not being a simple fix but, in fact, a new feature, this would go into 0.7, and 0.7 will have to support New Cerberus.

Actually, would you be willing work Cerberus 0.10 (soon to be renamed 1.0) support? I could really use some help there.

dkellner:
OK, then let's close this and focus on Cerberus 0.10/1.0 and Eve 0.7. I'm going to work on the Cerberus part first and maybe look into Eve integration after that. Unfortunately, I can't guarantee to make it as I'm going to travel for some (longer :)) time - if I start working on it, I'll comment in #776.

Issue #814, opened 2016-02-01
OPLOG DELETE


cbocta:
I don't know if any of you use OPLOG, but I have noticed that when I (soft) delete a document, the updated field is not the time of the deletion but the time of the previous last update of the doc.

When I look at the code oplog push is called on the 'original' doc instead of the 'marked' doc.
Can anyone explain me this behavior?

PR #816, opened 2016-02-01
Fix issue #814

cbocta:
If #814 is confirmed, here is a starting proposal
Signed-off-by: cbonnard <cyril.bonnard>


nicolaiarocci:
I added a test and it is now merged, thanks!

cbocta:
Great! Thx too!

PR #833, opened 2016-03-11
Make app installed as console-script not fail

txomon:
So I am doing all my python apps be deployed with pip install, and when using by console-scripts to launch servers, without this introspection, path will be wrong.

I really thought I had submitted this patch. It is really hacky, and maybe expensive, but just on the initialization.

### DIFF
--- a/eve/flaskapp.py
+++ b/eve/flaskapp.py
@@ -11,7 +11,6 @@
 """
 import inspect
 import os
-import sys
 
 import copy
 from events import Events
@@ -231,7 +230,7 @@ class Eve(Flask, Events):
                 abspath = inspect.getabsfile(
                     inspect.currentframe().f_back.f_back)
                 absdir = os.path.dirname(abspath)
-                pyfile = os.path.join(abspath, self.settings)
+                pyfile = os.path.join(absdir, self.settings)
             try:
                 self.config.from_pyfile(pyfile)
             except IOError:
--- a/eve/flaskapp.py
+++ b/eve/flaskapp.py
@@ -9,6 +9,7 @@
     :copyright: (c) 2016 by Nicola Iarocci.
     :license: BSD, see LICENSE for more details.
 """
+import inspect
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


txomon:
Probably we should be using https://github.com/mitsuhiko/flask/blob/master/flask/helpers.py#L777-L901 or something similar...

julianhille:
Hey can you be more specific in what you are doing exactly? 
This code looks really hacky to me and i also install everything with pip and start the servers with the cli and don't have these issues at all.

Are you using virtualenvs? What python version?
Please be as specific as possible maybe start with the installation and end with starting the server.
thanks

txomon:
@julianhille Just install your software through pip install zipfile, using console-script to launch the server.

You should see this problem if you run your application with console-script.

Also, although is a little hacky, it has the same result as the linked flask helper functions.

julianhille:
Sorry im unable to help as i don't understand what you are doing and i don't see the issue.

txomon:
The current code assumes that first argument represents a path to the main file, which is next to the settings.py. Any situation where this is not true, will break the Eve finding settings.py.

For example, when running python functions through [console-script](http://python-packaging.readthedocs.org/en/latest/command-line-scripts.html).

Please tell me which part I am not explaining correctly.

julianhille:
I was missing that you are using the console-script from setup.cfg and always thought about just a command through bash, my bad.
But your code assumes that two levels up is the entry point for the script which isn't correct either.
I don't have that issue, because i don't rely on the path creation for the default settings.py from eve and rather use my own settings creation and put it in as a dict.

i personally would remove that part 

            else:
                abspath = inspect.getabsfile(
                    inspect.currentframe().f_back.f_back)
                absdir = os.path.dirname(abspath)
                pyfile = os.path.join(absdir, self.settings)

completely and leave it to the developer to give the right file path or dictionary.

txomon:
The default is to take the file from the same dir as the script. This is inherited from flask, using the helper functions I put.

I also disagree with having the developer provide that. It's a 0 configuration framework, and should remain like that.

Also, that two backs are because the parent frame is in the same file, and is in the next frame where the app is.

nicolaiarocci:
This seems unnecessary and overcomplicated for most use cases. I am going to dodge it for the time being. If there are more requests in the future, we might look back at this.

PR #844, opened 2016-03-26
Closes #840: RegEx matches auth token from header

### LINKED ISSUE #840: TokenAuth prevents base64 decoding of tokens
Hi, I left a comment in the code [here](https://github.com/nicolaiarocci/eve/commit/24b30bc49dd8d5bb0017e8477910837bb218e802#diff-7303d4089a7f42fde94cc846a0c7d679R253).

> When I have a token that looks like `Authorization: token SGVsbG8=`, the a base64 decoder will give me `Hello`, but when `lower()` is applied to my token, I will get `²ìn`.
>
> To make TokenAuth work in my app, I override the `authorized` method and remove the `lower()` bit. I guess I'm wondering if this was the intended behavior?
>
> Cheers!

If it is the intended behavior, would it be okay if I made a PR to update the documentation to point out how to override the `authorized` method appropriately?

Thanks for this awesome project! :smile:

artburkart:
This PR introduces a test library called
[nose-parameterized](https://github.com/wolever/nose-parameterized),
which enables pretty flexible parameterized testing. This enabled me to
ensure the new `extract_token` method I wrote for the TokenAuth class
works as expected. I didn't want any unaccounted edge cases.

The `extract_token` method uses the following RegEx to grab the token
from the headers:

```python
match = re.search('token[\s]+(.*)', auth, re.IGNORECASE)
```

It matches anything beginning with the word "token" where the case of its
letters are ignored, followed by 1 or more whitespace characters, followed
by the actual token.

If you'd prefer for the behavior to simply be rolled back to what it
was before `lower()` was added, I am happy to accommodate that as well.

Let me know what you think! Cheers :smile:

PS: I just noticed my autoformatter rearranged the Python import
statements. If this is a problem, I am happy to undo that.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,6 +1,7 @@
 docutils==0.12
 flake8==2.3.0
 mccabe==0.3
+nose-parameterized==0.5.0
 pep8==1.5.7
 pip-tools==0.3.5
 pip-review==0.4
--- a/eve/auth.py
+++ b/eve/auth.py
@@ -9,7 +9,13 @@
     :copyright: (c) 2016 by Nicola Iarocci.
     :license: BSD, see LICENSE for more details.
 """
-from flask import request, Response, current_app as app, g, abort
+import re
+
+from flask import Response
+from flask import abort
+from flask import current_app as app
+from flask import g
+from flask import request
 from functools import wraps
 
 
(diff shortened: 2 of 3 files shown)


artburkart:
welp, trying to figure out where I went wrong

nicolaiarocci:
Hello, thanks for your contribution. Not sure what's going on on your side. I tried a pip install from develop branch and worked fine. 

In order to get the test suite to run, you need to add `nose-parameterized` to `test_requires` in setup.py. I didn't know about that tool, and it is pretty nice. However I am not sure we want to add a new dependency just for a single test. Maybe we will add it later, when we get to review/refactor the whole test suite, something that is already in the long term plan.

Also, using a regex seems a little overkill here, mainly because of the extra overhead. Of course you are trying to be coherent with the current behaviour introduced with `lower()`, but I feel that simply testing if the header starts with `Token` or `token` will be fine. If someone sends other variations (`toKEn`) we can just reject them.

If you are willing to resubmit with no additional dependencies and no regex, I'll gladly take your new PR.

artburkart:
I am happy to submit a new PR. 

One thing I want to point out, however, is that it's not a new dependency for a single test. It's a new dependency for 9 tests. The alternative is I write the same test 9 times with different input.

And a question: is there a specification somewhere that says `Token` and `token` should be acceptable inputs? I've only seen `token`.

Thanks for reviewing my PR! I'll get a new one to you soon :smile:

nicolaiarocci:
> And a question: is there a specification somewhere that says Token and token should be acceptable inputs? I've only seen token.

If you do a search for "token auth header" almost all of the results will show you a code snippet where Token is used ([this](http://www.django-rest-framework.org/api-guide/authentication/) is Django auth docs for example). I think it is just a common use case, probably due to the "Basic" header used with BasicAuth.

PR #851, opened 2016-04-10
Support etag cache validation for versioned documents 

nckpark:
- Support weak ETags, commonly applied by servers transmitting gzipped content.
- Allow old versions to be cache validated using ETags.
Without 'If-None-Match' style cache validation supported, data migrations
that update old versions and correctly compute new ETags, but don't change
updated_at dates, will appear to be valid in client caches even though the
data has been changed.


nicolaiarocci:
Thanks!

PR #862, opened 2016-05-15
Support ENFORCE_IF_MATCH flag

### LINKED ISSUE #657: Thoughts on If-Match header support
I don't really understand why the `IF_MATCH` option is so binary. Either you enable it and all your edit calls must be conditional or you have to disable it altogether. That is so annoying, especially when you are developing or testing the API, that I think the first thing most people do (as I did) is to disable this feature.

HTTP never require the If-Match header, even when Etag header is returned and I think it should be the same in Eve by default when `IF_MATCH` is enabled. You may have a `FORCE_IF_MATCH` configuration option if you want to enforce the use of conditional request though.

Also, the `Etag` shouldn't be in the payload but returned as a ETag header. If the client is not able to read that header, he won't be able to set the `If-Match` header anyway.

artburkart:
If `ENFORCE_IF_MATCH` is disabled and `IF_MATCH` is enabled, then
the client can choose whether to provide the `If-Match` header in
its request. When `If-Match` is provided, then the request is treated as
conditional, meaning the ETag must match. When `If-Match` is absent,
then the ETag check is not enforced.

The `ENFORCE_IF_MATCH` flag defaults to `True`

Closes #657

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -13,6 +13,7 @@ Patches and Contributions
 - Andrés Martano
 - Antonio Lourenco
 - Arnau Orriols
+- Arthur Burkart
 - Ashley Roach
 - Ben Demaree
 - Bjorn Andersson
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,13 @@ In Development
 Version 0.7
 ~~~~~~~~~~~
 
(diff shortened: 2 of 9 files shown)


nicolaiarocci:
👍

nicolaiarocci:
Excellent, thanks!

PR #864, opened 2016-05-19
Update endpoints.py

microhuang:
error

### DIFF
--- a/eve/endpoints.py
+++ b/eve/endpoints.py
@@ -156,7 +156,7 @@ def error_endpoint(error):
     .. versionadded:: 0.4
     """
     headers = None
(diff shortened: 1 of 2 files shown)


nicolaiarocci:
Hello and thanks for this. Could you please tell me more about the specific error/use case you are addressing with this?

microhuang:
Verification is not passed, get an error "'Unauthorized'object has no attribute'response' AttributeError:", "error" is "werkzeug.exceptions.Unauthorized", it does not have the response attribute

nicolaiarocci:
Could you please revert aadc49e so I can merge the original commit?

PR #887, opened 2016-08-03
Fix notifications example, create settings.

stasfilin:
I think we need change examples, 'cause we don't show how to use SETTINGS, and when user start notifications example he got error


nicolaiarocci:
Hello, you probably wanted to submit this to the eve-sqlalchemy repo?

PR #890, opened 2016-08-04
Fix notifications example

stasfilin:
The last version is incorrect, in this version I use mongo for example and create a simple scheme for test table.
I think we need update, 'cause when I start to learn Eve first - I start example and example doesn't work, and I've been confused

### DIFF
--- a/examples/notifications.py
+++ b/examples/notifications.py
@@ -19,9 +19,9 @@ from __future__ import print_function
 """
 from flask import request
 from eve import Eve
-from settings_notifications import SETTINGS, SQL, ValidatorSQL, Base
+from settings_notifications import SETTINGS
 
-app = Eve(auth=None, settings=SETTINGS, validator=ValidatorSQL, data=SQL)
+app = Eve(auth=None, settings=SETTINGS)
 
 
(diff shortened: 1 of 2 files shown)


nicolaiarocci:
Hello, thanks for this. What exactly was the problem that you experienced? I just ran the notifications example without issues (send a GET request to `localhost:5000`).

stasfilin:
@nicolaiarocci When i start example (not my) i got error; Need DOMAIN

nicolaiarocci:
Ah, right. I stupidly ran it with a settings.py file in the local folder 😄

stasfilin:
@nicolaiarocci yes, and if my example okay i can change all examples

nicolaiarocci:
I merged after a rebase (see https://github.com/nicolaiarocci/eve/commit/68c242a1e50ba02d6a58e38b854ff96f245a3a31). Also I added some minor refactoring/cleanup. Thanks!

stasfilin:
@nicolaiarocci so, I need to change another examples?

nicolaiarocci:
yes please

PR #892, opened 2016-08-04
Clarify Redis is an optional dependency.

### LINKED ISSUE #889: Docs are not clear about installation requirements
The docs say:

> Eve is powered by Flask, Redis, Cerberus, Events

but it does not indicate if all of those are required.

Specifically, I have failed to find anywhere in the docs if Redis is an optional dependency.

Looking into the requirements.txt and reading usage samples of Redis, `app = Eve()`  vs `app = Eve(redis=r)` also suggests Redis is optional.

However, that is too many hoops for those new to Eve to conclude about Redis requirement - many may give up having no option to host Redis. For example, PythonAnywhere users: [I couldn't tell from their docs whether redis is a hard requirement or something that you can use for some features](https://www.pythonanywhere.com/forums/topic/3730/#id_post_18968)).

mloskot:
Closes #889

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -8,7 +8,8 @@ human beings. It allows to effortlessly build and deploy highly customizable,
 fully featured RESTful Web Services.
 
 Eve is powered by Flask_, Redis_, Cerberus_, Events_ and offers support for
-both MongoDB_ and SQL backends [*]_.
+both MongoDB_ and SQL data storage backends [*]_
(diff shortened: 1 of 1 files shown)


nicolaiarocci:
Thanks! I went a little bit further and removed reds from the 'powered by' headline since it is already mentioned as a dependency for the rate-limiting feature.

mloskot:
Good idea.

Issue #810, opened 2016-01-16
post_internal fails when using URL_PREFIX or API_VERSION


daviskirk:
When setting `URL_PREFIX` option to a non-null value, the `post_internal` fails with a `TypeError`.
I first tried this on my own API but rechecked it in `eve-demo` which yields the error and traceback:

```
Traceback (most recent call last):
  File "run.py", line 35, in <module>
    a =  post_internal('people', {'firstname': 'Test', 'lastname': 'Name'})
  File "env/lib/python3.5/site-packages/eve/methods/post.py", line 247, in post_internal
    ids = app.data.insert(resource, documents)
  File "env/lib/python3.5/site-packages/eve/io/mongo/mongo.py", line 386, in insert
    datasource, _, _, _ = self._datasource_ex(resource)
  File "env/lib/python3.5/site-packages/eve/io/base.py", line 434, in _datasource_ex
    auth_field, request_auth_value = auth_field_and_value(resource)
  File "env/lib/python3.5/site-packages/eve/auth.py", line 258, in auth_field_and_value
    if '|resource' in request.endpoint:
TypeError: argument of type 'NoneType' is not iterable
```

To reproduce, add 

``` python
URL_PREFIX = 'test_api'
```

to `settings.py` and add

``` python
with app.test_request_context():
    post_internal('people', {'firstname': 'Test', 'lastname': 'Name'})
```

after `app = Eve()` in `run.py`.

It seems that `flask.request` is set to `None` in this case but up until now I can't figure out why.

nicolaiarocci:
Hello, thanks for reporting this.

1. With which Eve version is this happening? 
2. Which auth are you using, if any? 

I tried reproducing by adding the following test:

```python
def test_post_internal_with_url_prefix_or_api_version(self):
    self.app.config['URL_PREFIX'] = 'test_api'

    with self.app.test_request_context():
        post_internal(self.known_resource, {'firstname': 'Test', 'lastname': 'Name'})                               
```

Test however passes with BasicAuth and also with no auth at all.

daviskirk:
Hi,
Thanks for the quick answer!
I originally tested it with a stable Eve version 0.6.1.  However, I just pulled the master branch from github (0.6.2.dev0) and the result/error stays the same.  
I've also tested python versions is 3.5 and 3.4.  On clean python environments.
pip freeze shows:

```
Cerberus==0.9.2
Eve==0.6.1
Events==0.2.1
Flask==0.10.1
[...]

PR #893, opened 2016-08-05
Added unit test for issue 810.

sybrenstuvel:
The problem with #810 seems to be that code called by post_internal() relies on the Flask `request` object, even though the current request may not be a POST to the given resource. I'm not sure how to fix this myself, though.

### DIFF
--- a/eve/methods/common.py
+++ b/eve/methods/common.py
@@ -1040,6 +1040,8 @@ def resource_link():
     """
     path = request.path.strip('/')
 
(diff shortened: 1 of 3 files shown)


sybrenstuvel:
PS: Travis should fail on this, as I've only exposed the bug in a unit test, and not fixed the bug.

nicolaiarocci:
Thanks for this. It finally got merged, and you spared me some work 👍

Issue #915, opened 2016-09-18
List sent as 'multipart/form-data' does NOT pass validation step


nam4dev:
When sending list(s) in form data from the Client, they are not well-decoded from the backend side.
Actually, to get list(s) from `Flask.Request.form` Immutable Dict the method `get_lists(<form-data-key>)` shall be invoked.

This could be problematic as one shall send a list of files, file cannot be JSON serialized, not choice but to use form data...

For more details see stack overflow question [HERE](http://stackoverflow.com/questions/39516884/how-to-post-a-list-of-embedded-resource-ids-and-media-resources-files-at-the-s).

**Proposed PR: _#916_**

Leonidaz0r:
I am writing a solution to this currently, see #933. Would that fix your problem?

nicolaiarocci:
Looks like #933 could potentially fix both this ticket and #932.

PR #916, opened 2016-09-18
Fix validation for multipart/form-data Lists #915

nam4dev:
Hi Nicola,

First, I've seen your work about `MULTIPART_FORM_FIELDS_AS_JSON` option and that's great!
In the same intention, this PR to allow one who follows convention about `multipart/form-data` list type to send its data in a way Eve schema validation step will understand.

I hope I did not forget any Use Case, your feedback is welcomed :)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -91,6 +91,7 @@ Patches and Contributions
 - Mayur Dhamanwala
 - Mikael Berg
 - Mugur Rus
+- Namgyal Brisson
 - Nathan Reynolds
 - Niall Donegan
 - Nick Park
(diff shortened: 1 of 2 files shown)


nam4dev:
Hi,

**I agree with you, I'll rename it and will do UT for this use-case.**

> But I get to think about `form-data` problematic, what about nested dictionary containing a list of files sent as `form-data` ?

**I mean that one shall implement a generic layer to convert `form-data` format into well-understood python type (dict) to be properly evaluated by Validation step.**

> At first sight, list and dictionary (simple & nested) imbrications are possible, if a (list of) file(s) is in it then the Client side shall send it as form-data if one wants to process only 1 Request.

**Otherwise, another solution could be to send a (list of) file(s) as `base64 encoded string(s)` into `JSON` payload (using or not `MULTIPART_FORM_FIELDS_AS_JSON` option according to needs), which solves the issue pretty much without effort...**

*Please, give me your feedback about this*.

PR #917, opened 2016-09-26
Support for Cerberus 1.0 (work in progress)

dkellner:
Integrating Cerberus 1.0 was pretty straightforward and fun :). Most information is already in the commit message:

```
"Work in progress" because we still have to wait for the Cerberus issue
"`readonly` conflicts with `default`" to be resolved.
(see https://github.com/nicolaiarocci/cerberus/issues/268)

This is a rather big change. I still decided to do a single commit, as
intermediate commits would be in a non-working state anyway.

Breaking changes:

- `keyschema` was renamed to `valueschema` and `propertyschema` to
  `keyschema` (following changes in Cerberus).
- A PATCH on a document which misses a field having a default value will
  now result in setting this value, even if the field was not provided
  in the PATCH's payload.
- Error messages for `keyschema` are now returned as dictionary.
  Before: {'propertyschema_dict': 'propertyschema_dict'}
  Now: {'keyschema_dict': {'AAA': "value does not match regex '[a-z]+'"}}
- Error messages for `type` validations are different now (following
  changes in Cerberus).
- It is no longer valid to have a field with `default` = None and
  `nullable` = False.
  (see patch.py:test_patch_nested_document_nullable_missing)

In a nutshell, changes to the codebase are as follows:

- Add data layer independent subclass of `cerberus.Validator`
  * Support new signature of `__init__` and `validate`
  * Use `_config`-aware properties instead of bare member attributes to
    pass the `resource`, `document_id` and `persisted_document` to make
    them available to child validators
  * Add schema-docstrings to all `_validate_*` methods

- Adjust Mongo-specific `Validator` subclass
  * Adjust `_validate_type_*` methods (following changes in Cerberus)
  * Add schema-docstrings to all `_validate_*` methods

- Add custom ErrorHandler to support `VALIDATION_ERROR_AS_LIST`

- A few renames:
  * `ValidationError` -> `DocumentError`
  * `propertyschema` -> `keyschema` and `keyschema` -> `valueschema`

- Adjust tests due to different validation error messages
  (mostly for `type`)

- Remove `transparent_schema_rules` without replacement
- Remove `default`-handling, as Cerberus takes care of this now
```

I have two questions for the "final touch":
1. Am I supposed to add `..versionchanged` (etc.) comments myself? If yes, which version should I use? 0.7?
2. I removed the docstrings of `_validate_*`-methods because Cerberus now uses the docstring for schema validation. Should I re-insert them as comments? Most of them were pretty meaningless, though... (e.g. "Enable validation for `media` data type")


nicolaiarocci:
I'll review this later but at quick glance this looks like stella work, thank you!

1. My plan is to release v0.7 with old cerberus. This is to let people take advantage of Mongo Aggregation right away, with no need to deal with breaking changes. v0.8 would then be released, (relatively) shortly after v0.7. Of course, v0.8 would bit thing would be Cerberus 1.0.
2. Docstrings for validate methods we can loose of them.
3. Would be great if you could also update the documentation (`validation.rst` comes to mind, maybe other files too).

Again, great job.

dkellner:
> Would be great if you could also update the documentation (validation.rst comes to mind, maybe other files too).

Ah yes, of course :). I will take care of the Cerberus issue https://github.com/nicolaiarocci/cerberus/issues/268 first and then update documentation on both projects.

nicolaiarocci:
I deleted the `develop` branch so the PR got auto-closed. Could you resubmit against `master`? - Thanks!

dkellner:
Of course. I will do that later. There's still some work to do on this, but I have not forgot it. And now that we have Cerberus 1.1 (including the fix for the default & readonly issue) I can continue to work based on that.

PR #919, opened 2016-10-03
Escape the resource title in the XML render of the root doc

klambrec:
In the XML render of the root doc we are already escaping the href, but not the resource title. This patch adds escaping to the resource title.

This issue is particularly easy to hit when setting up Sub Resources as by default the the resource_title = url which shall contain characters that need to be escaped in XML.


nicolaiarocci:
Merged after rebase at: https://github.com/nicolaiarocci/eve/commit/b00c415a6458f4fc19dbd96bb9f8f6238440d6fa

Thanks!

PR #923, opened 2016-10-16
Add Python 3.5 as a supported interpreter

mattiaslundberg:
Add Python 3.5 as a supported interpreter in documentation and enables travis running of tests on Python 3.5.

To get everything working and upgrade of tox was needed in the dev requirements (travis already installed the latest version).

To get the tests running locally on my machine (Arch linux running Python 3.5.2) I needed to modify handling of attempts to modify immutables in `eve/io/mongo/mongo.py`. I suspect that this is the version of MongoDB used locally (3.2.9), but not sure about why I need to do the change (tests are passing in all other Python versions).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,24 @@
 language: python
-python: 2.7
-env:
-  - TOX_ENV=py26
-  - TOX_ENV=py27
-  - TOX_ENV=py33
-  - TOX_ENV=py34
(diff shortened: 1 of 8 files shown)


nicolaiarocci:
Merged after rebase: https://github.com/nicolaiarocci/eve/commit/3f31bc09efccece463b21c940d43f6ff96d9e790

Thanks!

PR #936, opened 2016-11-07
Feature: JSON_INDENT for pretty prints

omani:
If the user wants to see pretty printing in eg. the browser or console (curl) give him that option through settings.py


omani:
wrong branch. closed.

ddurieux:
Not sure it's the best way, I think it's better to pass an argument in URL to want display pretty JSON

omani:
please see #937

PR #938, opened 2016-11-07
Feature: Enable pretty printing of json response

omani:
If the user provides a query string named 'pretty' let the json be pretty by indenting the json.

### DIFF
--- a/eve/render.py
+++ b/eve/render.py
@@ -270,7 +270,12 @@ def render_json(data):
     .. versionchanged:: 0.1.0
        Support for optional HATEOAS.
     """
-    return json.dumps(data, cls=app.data.json_encoder_class,
+    set_indent = None
+
+    # make pretty prints available
+    if 'pretty' in request.args:
+        set_indent = 4
+    return json.dumps(data, indent=set_indent, cls=app.data.json_encoder_class,
                       sort_keys=config.JSON_SORT_KEYS)
 
 
--- a/eve/render.py
+++ b/eve/render.py
@@ -273,7 +273,7 @@ def render_json(data):
     set_indent = None
 
     # make pretty prints available
-    if 'pretty' in request.args:
+    if 'GET' in request.method and 'pretty' in request.args:
         set_indent = 4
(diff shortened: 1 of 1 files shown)


omani:
since ddurieux suggested:
> Not sure it's the best way, I think it's better to pass an argument in URL to want display pretty JSON

I changed query args instead of settings the option in settings.py

nicolaiarocci:
Hello thanks for this. A few notes:

- Needs to be documented or nobody will know about it. Since it's not a config setting I guess a good candidate would be the Filtering section of the Features page.
- Needs a proper test
- What is the use case for this, in the context of a REST API? Isn't the json going to be processed anyway?

omani:
Hi @nicolaiarocci,

* I will document this in features.rst
* I will provide a test unit for this (I first wanted to know if it is acceptable)
* the develop branch breaks my API with some errors I can't remember anymore. I think it was a missing key in app.config:

`
 # flask-pymongo compatibility

        self.config['MONGO_CONNECT'] = self.config['MONGO_OPTIONS'].get(
            'connect', True
        )
`

I think key MONGO_OPTIONS was missing in self.config. But maybe I just missed a change from v0.6.4 to v0.6.5.


* use case for this: if you want to test something in cli or browser you can just add a trailing &pretty and it will come out in a nice formatted way. I know there are tools like postman doing this per default. but I just wanted to have a way where I can do this by hand.
And yes the json is going to be processed anyway. So one could or should disable the pretty printing to avoid suffering from more overhead than needed. that's why I did it as a query arg rather than enable/disable this in the settings.py which would require a reload/restart of the API.

nicolaiarocci:
Ok I'm fine with merging once we have tests and docs updated. If there's an issue with `MONGO_OPTIONS` we can fix it. But please, let me merge from `develop`. Thanks!

nicolaiarocci:
@omani are you still planning to work on this, or should I close the PR?

omani:
hi @nicolaiarocci,

you can close this PR. I will create a complete PR for the develop branch.

thanks.

omani:
please see #946.

PR #948, opened 2016-12-11
Added serialization rule for boolean

### LINKED ISSUE #947: Serializer for boolean
If I use a field, where the type is "integer", I can also post a string containing an integer and it will be converted by the serializer.
Is there a reason this does not happen for booleans?
It is missing in this list:

https://github.com/nicolaiarocci/eve/blob/develop/eve/io/mongo/mongo.py#L77

NotSpecial:
A range of values is now serialized to boolean, namely:

0, 1, '0', '1', 'false', 'False', 'true', 'True'

This covers the most common boolean inputs.

Closes #947

### DIFF
--- a/eve/io/mongo/mongo.py
+++ b/eve/io/mongo/mongo.py
@@ -80,6 +80,8 @@ class Mongo(DataLayer):
         'integer': lambda value: int(value) if value is not None else None,
         'float': lambda value: float(value) if value is not None else None,
         'number': lambda val: json.loads(val) if val is not None else None,
+        'boolean': lambda v:
+        {'1': True, 'true': True, '0': False, 'false': False}[str(v).lower()],
         'dbref': lambda value:
         DBRef(value['$col'], value['$id'], value['$db']
               if '$db' in value else None) if value is not None else None,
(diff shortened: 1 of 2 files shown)


eve/io/mongo/mongo.py
```diff
@@ -80,6 +80,8 @@ class Mongo(DataLayer):
         'integer': lambda value: int(value) if value is not None else None,
         'float': lambda value: float(value) if value is not None else None,
         'number': lambda val: json.loads(val) if val is not None else None,
+        'boolean': lambda v:
+        {'1': True, 'true': True, '0': False, 'false': False}[str(v).lower()],
```
Leonidaz0r: 1. We could also include things like "yes", "no", "on", "off", etc... But maybe it is overkill.
2. I would do something like {...}.get(str(v).lower()). That way we explicitly return None, if the value is not in that list. Currently we achieve the same thing indirectly by creating a KeyError, which then generates a None somewhere else. If it is explicit, there is less confusion when reading the code.

Leonidaz0r: I see. So the solution seems good.

nicolaiarocci:
Thanks!

PR #951, opened 2016-12-19
Current stable mongodb version is 3.4

gmargari:
Fix bug that occured when (trying to) modify the _id of object.
Used to give a warning, now gives internal error because the mongodb version is not in the list above.


nicolaiarocci:
Awsome, thank you!

Issue #918, opened 2016-09-27
DOMAIN dictionary missing or wrong with gunicorn


amoxyc:
Hello!

Everything works fine on Heroku with Procfile 
`web: python run.py`

But when I switch to gunicorn, exception appears:

>  [2016-09-27 19:27:21 +0000] [9] [ERROR] Exception in worker process
> 2016-09-27T19:27:21.633227+00:00 app[web.1]: Traceback (most recent call last):
> 2016-09-27T19:27:21.633228+00:00 app[web.1]:   File "/app/.heroku/python/lib/python2.7/site-packages/gunicorn/arbiter.py", line 557, in spawn_worker
> 2016-09-27T19:27:21.633229+00:00 app[web.1]:     worker.init_process()
> 2016-09-27T19:27:21.633229+00:00 app[web.1]:   File "/app/.heroku/python/lib/python2.7/site-packages/gunicorn/workers/base.py", line 126, in init_process
> 2016-09-27T19:27:21.633230+00:00 app[web.1]:     self.load_wsgi()
> 2016-09-27T19:27:21.633231+00:00 app[web.1]:   File "/app/.heroku/python/lib/python2.7/site-packages/gunicorn/workers/base.py", line 136, in load_wsgi
> 2016-09-27T19:27:21.633232+00:00 app[web.1]:     self.wsgi = self.app.wsgi()
> 2016-09-27T19:27:21.633232+00:00 app[web.1]:   File "/app/.heroku/python/lib/python2.7/site-packages/gunicorn/app/base.py", line 67, in wsgi
> 2016-09-27T19:27:21.633233+00:00 app[web.1]:     self.callable = self.load()
> 2016-09-27T19:27:21.633234+00:00 app[web.1]:   File "/app/.heroku/python/lib/python2.7/site-packages/gunicorn/app/wsgiapp.py", line 65, in load
> 2016-09-27T19:27:21.633235+00:00 app[web.1]:     return self.load_wsgiapp()
> 2016-09-27T19:27:21.633235+00:00 app[web.1]:   File "/app/.heroku/python/lib/python2.7/site-packages/gunicorn/app/wsgiapp.py", line 52, in load_wsgiapp
> 2016-09-27T19:27:21.633236+00:00 app[web.1]:     return util.import_app(self.app_uri)
[...]

Issue #820, opened 2016-02-08
Problems with relative path to settings.py and a set EVE_SETTINGS environment var


britalmeida:
Hello,
I have recently run into a problem where I set EVE_SETTINGS with a (correct) relative path.

I am reading this var in my application code and making it absolute before passing it to  Eve(settings=...
I do this as a workaround to a issue similar to #137 where my relative path ends up being relative to something else that my initial running directory. (In my case it gets appended to ~/.pyenv/versions/venv-name/lib/python2.7/site-packages/eve/ ).

My specific problem is that eve is ignoring the path I give it by overriding it with the value from the environment var in 'load_config'. This is not the behaviour that I expected, I would think that the argument had overriding precedence over the env var. Maybe the documentation could be clarified here?
Still, I am left with no nice workaround except for not using EVE_SETTINGS at all and providing my own var that I can process before passing to eve.

Issue #713, opened 2015-09-15
Handle exceptions when parsing settings.py


mandarvaze:
I had a bad class/SQLAlchemy model. All I get is very generic ``ConfigException: DOMAIN dictionary missing or wrong. `` which is not at all helpful.

I had to put a trace and debug the code, when the following was printed on the debugger console.

    sqlalchemy.exc.NoReferencedTableError: Foreign key associated with column 'xxxxx' could not find table 'xxxx' with which to generate a foreign key to target column 'id'

There was a typo, which I was able to quickly fix based on the above error, but getting this error itself was impossible in "normal" way.

While I came across this when using eve-sqlalchemy (as may be evident from the exception) the problematic code is in the flaskapp.py of eve. 
(*I had almost written a bug against eve-sqlalchemy, when further digging revealed that problem is in the eve code*)

Here is a possible fix.
Either we can print, or raise the exception. But ``pass`` in the ``except`` clause doesn't help anyone.
I'm not sure how you want to handle this, so not sending PR yet. Let me know, and I can send in the PR

``flaskapp.py``

```diff
*** 211,222 ****
              else:
                  abspath = os.path.abspath(os.path.dirname(sys.argv[0]))
                  pyfile = os.path.join(abspath, self.settings)
              try:
                  self.config.from_pyfile(pyfile)
!             except Exception as e:
!                 raise e
                  # assume an envvar is going to be used exclusively
                  pass

          # overwrite settings with custom environment variable
          envvar = 'EVE_SETTINGS'
--- 211,221 ----
              else:
                  abspath = os.path.abspath(os.path.dirname(sys.argv[0]))
                  pyfile = os.path.join(abspath, self.settings)
              try:
                  self.config.from_pyfile(pyfile)
!             except:
                  # assume an envvar is going to be used exclusively
                  pass

          # overwrite settings with custom environment variable
          envvar = 'EVE_SETTINGS'


```

nicolaiarocci:
Hey thanks for this. Digging a little further would have showed that this has been [addressed](https://github.com/nicolaiarocci/eve/blob/develop/eve/flaskapp.py#L227-L233) already on development branch. Relevant CHANGES line:

    +- Fix: some schema errors were not being reported as SchemaError exceptions
    +  A more generic 'DOMAIN missing or wrong' message was returned instead.

mandarvaze:
Awesome.
Unfortunately I tested my proposed fix, only on my local setup, which had `pip install` ed version.
[...]

Issue #209, opened 2013-12-12
Fail to run nose tests: Eve try to load settings.py from dir of noserunner.py


clark0x:
My project is using nose as unit test framework. With eve, it raise error:

```
ERROR: Failure: IOError ([Errno 2] Unable to load configuration file (No such file or directory): '/Applications/PyCharm.app/helpers/pycharm/settings.py')
```

I have located this line in `flaskapp.py` (line 195, eve 0.2):

```
abspath = os.path.abspath(os.path.dirname(sys.argv[0]))
```

`sys.argv` has two elements:
1. `sys.argv[0]` is path of `noserunner.py`, which is in the dir in the error message above;
2. `sys.argv[1]` is path of tests dir.

I can alter `0` to `1` and provide  copy of `settings.py` in the tests dir to pass the nose tests, but this is not a good idea.

PR #962, opened 2017-01-14
better locating of settings.py

bomb-on:
First of all, sorry for putting this issue on the table again, but it seems that this is still not completely solved. Instead there are some very simple workarounds but I would like to see Eve even smarter than those workarounds (or at least properly documented issue), so here are my 2 cents...

# The Problem
I've been reading #137, #209, #713, #820, #918 (and others), along with issues related to this specific problem in [eve-sqlalchemy](https://github.com/RedTurtle/eve-sqlalchemy). Although some of those issues are outdated because they were related to older Eve versions, it seems that there are some isolated cases where this is still occurring.
In my case, I have a `settings.py` file in the root of my project, along with `application.py` where Eve is running. The application itself works fine in my local dev environment (OS X 10.12 with Python 3.5.1) and also when deploying it on AWS (AWS linux with Python 3.4) but I have a problem when I try to run tests with `py.test` on any platform because `settings.py` is never properly discovered.
We know already that the problem lies in [`load_config`](https://github.com/nicolaiarocci/eve/blob/develop/eve/flaskapp.py#L202) method, specifically few lines which are responsible for locating `settings.py` file:
```
if isinstance(self.settings, dict):
    self.config.update(self.settings)
else:
    if os.path.isabs(self.settings):
        pyfile = self.settings
    else:
        abspath = os.path.abspath(os.path.dirname(sys.argv[0]))  # <-- this!
        pyfile = os.path.join(abspath, self.settings)
    try:
        self.config.from_pyfile(pyfile)
    except IOError:
        # assume envvar is going to be used exclusively
        pass  # <--- why pass? what if envvar is not used in the end??
    except:
        raise
```
which in my case results with either `pytest` binary (`/Users/oxxo/.virtualenvs/evetest/bin/pytest`), or if I try to run it in PyCharm with its `helpers` folder (`/Applications/PyCharm.app/Contents/helpers/pycharm`).

# Proposed Solutions
There are already few possible (and very simple) solutions to this, but I think it might be difficult to other people to figure it out without actually searching the repository's issues, so my suggestions are:

- remove the "magic" behind discovering `settings.py` completely and "force" explicitly passing settings to Eve constructor, either as a dictionary or as an absolute path to `settings.py`. (IMHO, this would be the worst solution :))
- just raise a proper exception instead of `ConfigException('DOMAIN dictionary missing or wrong.')`. Seems like this is what others are tripping over the most. There was some mentioning of this in #713, but to me it looks like this is still not solved completely.
- improve the method of locating `settings.py`. This would keep Eve's "magic" on the track and in my opinion this is the best solution, therefore I went in that direction and it resulted with this pull request :)

Anyhow, no matter which solution will be applied, it should be **properly documented**, at least to avoid future tripping over :)

Thanks a lot for this great framework, keep up with good work and thanks a lot for your time! :)


bomb-on:
There are also 2 tiny additions to `.gitignore` and I fixed the typo in [`test_create_indexes()`](https://github.com/nicolaiarocci/eve/blob/develop/eve/tests/config.py#L490) which solves #960.

bomb-on:
Eh, well, seems that tests are failing when running with `tox` because tests folder is not in `sys.path` after all (as I thought)... Will try to figure this out and update PR...

nicolaiarocci:
I like the general approach you are taking here. Ayuto-magically figuring out a file path, however, is probably not so naive as it might seem at first glance.

bomb-on:
Indeed :)
Straight out of the head, we might add the path manually just for this test case if you agree, or I can try to figure out a bit "smarter" way to find the file in this particular case.
I'll try to play around with it a bit more and hopefully propose "fully" working solution :)

bomb-on:
So, this is passing now. I haven't test it on Windows because I have no Windows machine. Also, sorry for multiple commits :)

dkellner:
@bomb-on You could use `git rebase --interactive ...` to squash those multiple commits together and make the history nicer before Nicola will merge this :).

bomb-on:
@dkellner Oh, rebase? Damn, I did it once in my life and I have to admit that it didn't end up well :) But shouldn't this end up as a single commit after merging since my branch won't exist in this repo? Sorry, but I am still a bit noob-ish with git :)

About the documentation... Sure I could update it but I intentionally left it to @nicolaiarocci because I believe he knows the best what's happening behind the scenes and to transform it into human-readable sentence :)

nicolaiarocci:
@bomb-on you should update the docs too. Just need to update the [source rst file](https://github.com/nicolaiarocci/eve/blob/develop/docs/config.rst). After that is done, I will rebase and merge for you.

bomb-on:
@nicolaiarocci No probs! I'll try to do it sometimes today!

bomb-on:
@nicolaiarocci This is done. As usual, I tend to overcomplicate even the simplest explanations, so if that is a case here too, please let me know and I'll try to reformulate some of the stuff :)

nicolaiarocci:
Merged after rebase (see https://github.com/nicolaiarocci/eve/commit/d57176a7a9ec9068b59094b1b1e786723d667904)

Thanks!

Issue #967, opened 2017-01-19
Versioning doesn't work with U-RRA


klambrec:
The basic schema layout for this issue may look something like this.
```python
users = {
    'schema': {
        'username': {
            'type': 'string',
        },
        'password': {
            'type': 'string',
        },
    }
}

products = {
    'authentication': UserAuth,
    'auth_field': 'owner',
    'versioning': True,
    'schema': {
        'name': {
            'type': 'string',
            'maxlength': 300,
        }
    }
}


invoices = {
    'authentication': UserAuth,
    'auth_field': 'owner',
    'versioning': True,
    'schema': {
        'description': {
            'type': 'string',
            'maxlength': 300,
        },
        'products': {
            'type': 'list',
            'schema': {
                'type': 'dict',
                'schema': {
                    '_id': {
                        'type': 'objectid'
                    },
                    '_version': {
                        'type': 'integer'
                    }
                },
                'data_relation': {
                    'resource': 'products',
                    'field': '_id',
                    'embeddable': True,
                    'version': True
                }
            }
        },
    }
}

DOMAIN = {
    'users':
        users,
    'products':
        products,
    'invoices':
        invoices
}
```
A very basic UserAuth (obviously not secure)
```python
class UserAuth(BasicAuth):
    """ Authenticate based on username & password"""
    def check_auth(self, username, password, allowed_roles, resource, method):
        users = app.data.driver.db['users']
        lookup = {'username': username}
        user = users.find_one(lookup)
        self.set_request_auth_value(user['_id'])
        return True
```

After inserting some sample data and making a request for /invoices?embedded={"products": 1} a 404 is returned
`{'_error': {'message': "Unable to locate embedded documents for 'products'", 'code': 404}, '_status': 'ERR'}`

This works perfectly fine when removing either auth_field or versioning from products. 

Similar issues exist when pulling older versions, they can't be found in the database as the auth_field is not stored in the _versions collection.

Tested against PyPi release of Eve.

PR #970, opened 2017-01-20
Fix to push auth_field onto versioned documents

klambrec:
This patch is a proposed fix for #967.


nicolaiarocci:
Excellent. Would be great if you could also PR a test for this fix, so we don't fall into any regression in the future.

PR #979, opened 2017-02-04
Add snippet that implements a list of embedded items

jrc:
Add snippet that implements a list of items that supports both list-level and item-level CRUD operations. The goal was to make it possible to use a single GET to get the entire list (including all items) but also use a single POST to append an item (rather than PATCHing the list).

The solution was to database event hooks to inject the embedded child documents into the parent list before it's returned to the client (and also delete the child items when the parent list is deleted).


nicolaiarocci:
This was auto-closed because I deleted the `develop` branch. Could you resubmit against `master`? Thanks!

PR #990, opened 2017-02-20
fix create_index when using mongo_prefix and uris

petrjasek:
when there is `MONGO_URI` defined it will be used no matter
if the resource is using prefix or not. instead of custom
handling use pymongo helper.


nicolaiarocci:
Fails on Py26.

petrjasek:
ups, fixed now

nicolaiarocci:
Merged with rebase, see https://github.com/nicolaiarocci/eve/commit/d1054ea536fc6d34e271a5f74c95c26e729159a8

nicolaiarocci:
Thanks man!

PR #1001, opened 2017-03-21
Support for Cerberus 1.1

bcrochet:
(Based on original patch from dkellner)

This is a rather big change. I still decided to do a single commit, as
intermediate commits would be in a non-working state anyway.

Breaking changes:

- `keyschema` was renamed to `valueschema` and `propertyschema` to
  `keyschema` (following changes in Cerberus).
- A PATCH on a document which misses a field having a default value will
  now result in setting this value, even if the field was not provided
  in the PATCH's payload.
- Error messages for `keyschema` are now returned as dictionary.
  Before: {'propertyschema_dict': 'propertyschema_dict'}
  Now: {'keyschema_dict': {'AAA': "value does not match regex '[a-z]+'"}}
- Error messages for `type` validations are different now (following
  changes in Cerberus).
- It is no longer valid to have a field with `default` = None and
  `nullable` = False.
  (see patch.py:test_patch_nested_document_nullable_missing)

In a nutshell, changes to the codebase are as follows:

- Add data layer independent subclass of `cerberus.Validator`
  * Support new signature of `__init__` and `validate`
  * Use `_config`-aware properties instead of bare member attributes to
    pass the `resource`, `document_id` and `persisted_document` to make
    them available to child validators
  * Add schema-docstrings to all `_validate_*` methods

- Adjust Mongo-specific `Validator` subclass
  * Adjust `_validate_type_*` methods (following changes in Cerberus)
  * Add schema-docstrings to all `_validate_*` methods

- Add custom ErrorHandler to support `VALIDATION_ERROR_AS_LIST`

- A few renames:
  * `ValidationError` -> `DocumentError`
  * `propertyschema` -> `keyschema` and `keyschema` -> `valueschema`

- Adjust tests due to different validation error messages
  (mostly for `type`)

- Remove `transparent_schema_rules` without replacement
- Remove `default`-handling, as Cerberus takes care of this now


dkellner:
Great, thank you for re-opening this! I've quite busy with Eve-SQLAlchemy for now so did not have time myself to look into this again.

I just saw that `setup.py` and `tox.ini` still use `dependency_links` to install my development branch of Cerberus for testing - we should remove this. And AFAIK dependency link processing has been removed from pip.

nicolaiarocci:
Excellent. Thank you @bcrochet for taking up the torch. @dkellner I'd appreciate a thoughtful review when you have time (use the Review feature).

bcrochet:
@dkellner Removed the dependency_links lines from setup.py and tox.ini.

eve/validation.py
```diff
@@ -12,5 +12,144 @@
...
+        if not self.persisted_document or \
+           field not in self.persisted_document:
+            super(Validator, self)._normalize_default_setter(mapping, schema,
+                                                             field)
+
+    def _validate_dependencies(self, dependencies, field, value):
+        """ {'type': ['dict', 'hashable', 'hashables']} """
+        persisted = self._filter_persisted_fields_not_in_document(dependencies)
+        if persisted:
+            dcopy = copy.copy(self.document)
```
funkyfuture: would a deepcopy be less error prone here?

dkellner:
@nicolaiarocci I will review this during the next days.

In the meantime we can discuss about updating the documentation (as already mentioned here: https://github.com/pyeve/eve/pull/917#issuecomment-249575506). Additionally, we should add a section about upgrading, as this will introduce some breaking changes.

nicolaiarocci:
Yes docs need to be amended and upgrading should be covered.

eve/validation.py
```diff
@@ -12,5 +12,144 @@
...
+        else:
+            super(Validator, self)._validate_dependencies(dependencies, field,
+                                                          value)
+
+    def _filter_persisted_fields_not_in_document(self, fields):
+        def persisted_but_not_in_document(field):
+            return field not in self.document and \
+                self.persisted_document and \
+                field in self.persisted_document
+        return list(filter(persisted_but_not_in_document, fields))
```
dkellner: Agree, and it should make it more readable, too.

eve/validation.py
```diff
@@ -12,5 +12,144 @@
...
+    @property
+    def resource(self):
+        return self._config.get('resource', None)
+
+    @resource.setter
+    def resource(self, value):
+        self._config['resource'] = value
+
+    @property
+    def document_id(self):
```
dkellner: Just to clarify: This is nothing concerning this very PR, right?

eve/validation.py
```diff
@@ -12,5 +12,144 @@
...
+    @property
+    def persisted_document(self):
+        return self._config.get('persisted_document', None)
+
+    @persisted_document.setter
+    def persisted_document(self, value):
+        self._config['persisted_document'] = value
+
+
+class DefaultErrorHandler(cerberus.errors.BasicErrorHandler):
```
dkellner: I agree that we could do this check only one time at application startup. But I wonder if creating a dynamic type based on a config item can get confusing once someone will subclass this again.

How about renaming `DefaultErrorHandler` to something like `SingleErrorsAsStringErrorHandler` (or a better name if you can come with one) and then amend ```Validator.__init__``` as follows?
  ```python
  if not config.VALIDATION_ERROR_AS_LIST:
      kwargs['error_handler'] = SingleErrorsAsStringErrorHandler
  ```

But again I cannot see a "real" issue with your (definitely more sophisticated) suggestion.

eve/validation.py
dkellner: Yes, especially regarding the fact that the child validator can do normalization, too. I just looked into this and `copy.copy` was used in the original code as well. I think that's the reason I kept in there in the beginning. A "real" fix would include a test case where a plain copy fails so we never need to wonder about this again.

dkellner:
I looked into this again, but this is my code without changes (so far) - so I'm not sure if I'm really the right person to review it ;). Nonetheless, I took some notes regarding needed documentation changes. The list is probably not complete, but a hopefully a good start:

  - `_validate_type_uuid` in `custom_idfields.rst` is wrong now (signature + return value)
  - same for `_validate_type_objectid` in `validation.rst`
  - still contains references to 'propertyschema' (should be 'keyschema' now),
    this is probably true for other renames as well
  - remove references to `TRANSPARENT_SCHEMA_RULES` (config.rst, validation.rst)

@bcrochet Can you implement @funkyfuture's suggestions (amend to my commit if you want, I think that's fine in this case) and add another commit for the documentation changes?

bcrochet:
@dkellner You're absolutely right. It is mostly your original patch. I think I *maybe* changed 2 lines, just to get it working with the latest cerebrus. But yes, I will take @funkyfuture 's suggestions and amend this commit.

eve/validation.py
funkyfuture: that's correct.

funkyfuture:
actually, i'm not sure whether this errorhandler is really needed. in cerberus the argument was that current client code should already be ready to deal with such structures as they were used for 2+ errors.

nicolaiarocci:
@bcrochet is this work ready for merge, or you still have work left to do?

bcrochet:
@nicolaiarocci I'm currently rebasing and will have some changes, so it's not quite ready.

eve/validation.py
bcrochet: I'm leaving this for later. I'm not familiar enough with the code to be able to make an informed change here.

bcrochet:
I'd consider this ready. However, the python2.6 job is not passing, due to python2.6 being not found. Should this be dropped from tox.ini?

nicolaiarocci:
Looks like the latest Travis update broke our .travis.yml. It Can you please rebase (again) on top of master? Thanks!

bcrochet:
Rebase done. I also went ahead and included doc changes.

nicolaiarocci:
Excellent. One last effort and we're done. Please, update the 'breaking changes' section in the changelog. I guess a good reference would be Cerberus' own "upgrading to 1.0" page at https://cerberus.readthedocs.io/en/stable/upgrading.html.

bcrochet:
@nicolaiarocci Updated

nicolaiarocci:
Merged after rebase,see https://github.com/pyeve/eve/commit/3bfff772f30e6f3d83db58d8e4bbc5c02041e12a.

Stellar, thank you.

PR #1013, opened 2017-04-20
Internal resource, oplog enabled and *_internal raises keyerror

einarhuseby:
Using global configuration ```OPLOG=True```, resource definition ```'internal_resource': True``` and then performing an \*_internal (* methods defined in OPLOG_METHODS) will trigger a keyerror in ```oplog_push``` entry update since the internal resource is not in config.URLS ```'r': config.URLS[resource]``` (tested 0.7.1 and 0.7.2)

Changing the  check in oplog_push:
```if not config.OPLOG or op not in config.OPLOG_METHODS:``` 
to
```if not config.OPLOG or op not in config.OPLOG_METHODS or resource not in config.URLS:```
should make sense

### DIFF
--- a/eve/methods/common.py
+++ b/eve/methods/common.py
@@ -34,1153 +34,1147 @@ from werkzeug.datastructures import MultiDict, CombinedMultiDict
 
 
 def get_document(resource, concurrency_check, **lookup):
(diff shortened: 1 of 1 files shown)


nicolaiarocci:
Thanks!

Issue #1028, opened 2017-06-20
Bulk update error


Amedeo91:
Hi all,

if I set up the id_field to something different to the _id, Eve always replay with 404 as it is using the _id to perform the operation

Regards,
Amedeo

Issue #1010, opened 2017-04-10
POST/PATCH with multi ids


Amedeo91:
Hi all,

for my application I need to provide end-point as follow:
1 carts/<cart_id> --> datasource: carts
2 carts/<cart_id>/products/<product_id> --> datasource: products
3 carts/<cart_id>/products/<product_id>/details/<detail_id> --> datasource: details

My goal is to be able to do all the HTTP method (GET, PATCH, PUT, POST and the embedded resource solver) in using the parameter passed in the URL.

Note: <product_id> and <cart_id> are to be hidden from the client for details and only <cart_id> should be hidden for the products

Below some example:
- endpoint 1: GET/DELETE search_filter= {id:<cart_id>}
                     PUT/PATCH filter_to_update = {id:<cart_id>}
                     POST add the query parameter to the request body = {id:<cart_id>}
- endpoint 2: GET/DELETE search_filter= {cart_id:<cart_id>,id:<product_id>}
                     PUT/PATCH filter_to_update = {cart_id:<cart_id>,id:<product_id>}
                     POST add the query parameter to the request body = {cart_id:<cart_id>,id:<product_id>}
- endpoint 3: GET/DELETE search_filter= {cart_id:<cart_id>,product_id:<product_id>, id:<detail_id>}
                     PUT/PATCH filter_to_update = {cart_id:<cart_id>,product_id:<product_id>, id:<detail_id>}
                     POST add the query parameter to the request body = {cart_id:<cart_id>,product_id:<product_id>, id:<detail_id>}
 
At the moment I was able to meet my requirement by both adding some hook event (pre_POST, pre_PATCH, pre_PUT) and some customization in the dataLayer. 
However, I think it would be nicer to update Eve business logic (especially for the embedded  documents). 

Moreover, I reckon that the current behavior is kind-of buggy since I am able to POST a new element (i.e. a new product) and I am not able to GET it if the client does not explicitly provide the cartId at the product insert time.

The scenario above it is kind of usefull if the primary-key of our resource is made by more than only one field (therefore, {cart_id:<cart_id>,product_id:<product_id>, id:<detail_id>} should be a primary key for the details resource)

Could you please share with us what do you think about the above scenario?

Regards,
Amedeo

PR #1030, opened 2017-06-20
Bulk delete

Amedeo91:
Bulk delete fix +Adding callback on the bulk delete flow for customization purpose

Issue #1028 
Issue #1010 --> to support the possibility to perform a bulk delete with a multi key unique id (Note: only for not soft delete, as per soft delete is possible to achieve it with customisation in the datalayer)

### DIFF
--- a/eve/methods/delete.py
+++ b/eve/methods/delete.py
@@ -38,13 +38,14 @@ def deleteitem(resource, **lookup):
 
 
 def deleteitem_internal(
-        resource, concurrency_check=False, suppress_callbacks=False, **lookup):
+        resource, concurrency_check=False, suppress_callbacks=False, original=None, **lookup):
     """ Intended for internal delete calls, this method is not rate limited,
     authentication is not checked, pre-request events are not raised, and
     concurrency checking is optional. Deletes a resource item.
 
     :param resource: name of the resource to which the item(s) belong.
     :param concurrency_check: concurrency check switch (bool)
+    :param original: original document if already fetched from the database
     :param **lookup: item lookup query.
 
     .. versionchanged:: 0.6
@@ -83,7 +84,7 @@ def deleteitem_internal(
     """
     resource_def = config.DOMAIN[resource]
     soft_delete_enabled = resource_def['soft_delete']
-    original = get_document(resource, concurrency_check, **lookup)
+    original = get_document(resource, concurrency_check, original, **lookup)
     if not original or (soft_delete_enabled and
                         original.get(config.DELETED) is True):
         abort(404)
@@ -137,8 +138,7 @@ def deleteitem_internal(
             # get_document() call since it also deals with etag matching, which
             # is still needed. Also, this lookup should never fail.
             # TODO not happy with this hack. Not at all. Is there a better way?
-            original = app.data.find_one_raw(
-                resource, original[resource_def['id_field']])
+            original = app.data.find_one_raw(resource, **lookup)
 
         for field in media_fields:
             if field in original:
@@ -150,7 +150,7 @@ def deleteitem_internal(
                     app.media.delete(original[field], resource)
 
         id = original[resource_def['id_field']]
-        app.data.remove(resource, {resource_def['id_field']: id})
+        app.data.remove(resource, lookup)
 
         # TODO: should attempt to delete version collection even if setting is
         # off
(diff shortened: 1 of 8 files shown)


nicolaiarocci:
Are you basically attempting to make DELETE a batch operation, whereas so far it's been an individual (document endpoint) or complete (resource endpoint) operation?

eve/methods/delete.py
```diff
@@ -193,20 +193,34 @@ def delete(resource, **lookup):
...
     .. versionadded:: 0.0.2
     """
-    getattr(app, "on_delete_resource")(resource)
-    getattr(app, "on_delete_resource_%s" % resource)()
 
     resource_def = config.DOMAIN[resource]
+    getattr(app, "on_delete_resource")(resource)
+    getattr(app, "on_delete_resource_%s" % resource)()
+    default_request = ParsedRequest()
+    if config.DOMAIN[resource]['soft_delete']:
```
nicolaiarocci: could use `resource_def` here?

eve/methods/delete.py
```diff
@@ -193,20 +193,34 @@ def delete(resource, **lookup):
...
+        # get_document should always fetch soft deleted documents from the db
+        # callers must handle soft deleted documents
+        default_request.show_deleted = True
```
nicolaiarocci: Why returning already deleted documents to the callback? Busines logic should have been applied on original deletion, right?

Amedeo91:
Hi Nicola, basically, I just want to enforce that we should delete what we fetch. Therefore, if in the URL we are allowing to specify some parameters (i.e. person with for follow url users/<regex("[a-f0-9]{24}"):person>/invoices --> lookup = {"person" : value}), we should use the some parameters to do the delete.

eve/methods/delete.py
Amedeo91: right

eve/methods/delete.py
Amedeo91: Well, in this case, I would like to offer to the client possibility to customize what he wants to delete. A possible use case can be that in the collection there are documents which should not be deleted by the client (i.e. included service for a trip, like included car or included checked baggage).

Therefore, it might be possible by modifying the query (in the case on hard delete) or modify the originals (removing what we do not what to delete) in the callback.

eve/methods/delete.py
```diff
@@ -38,7 +38,7 @@ def deleteitem(resource, **lookup):
 
 
 def deleteitem_internal(
-        resource, concurrency_check=False, suppress_callbacks=False, **lookup):
+        resource, concurrency_check=False, suppress_callbacks=False, original=None, **lookup):
```
nicolaiarocci: missing docstring update here

eve/methods/delete.py
```diff
@@ -150,7 +149,8 @@ def deleteitem_internal(
                     app.media.delete(original[field], resource)
 
         id = original[resource_def['id_field']]
-        app.data.remove(resource, {resource_def['id_field']: id})
+        # We should delete what we fetch. Since the id_field might not be unique, let's use the lookup
```
nicolaiarocci: this comment can be deleted

eve/methods/delete.py
```diff
@@ -193,20 +193,34 @@ def delete(resource, **lookup):
...
-    getattr(app, "on_delete_resource_%s" % resource)()
 
     resource_def = config.DOMAIN[resource]
+    getattr(app, "on_delete_resource")(resource)
+    getattr(app, "on_delete_resource_%s" % resource)()
+    default_request = ParsedRequest()
+    if resource_def['soft_delete']:
+        # get_document should always fetch soft deleted documents from the db
+        # callers must handle soft deleted documents
+        default_request.show_deleted = True
```
nicolaiarocci: this behavior should probably be explained in the [Soft Delete](http://python-eve.org/features.html#soft-delete) section of the docs. Explain that the new callback will include soft-deleted documents when soft delete is active.

eve/tests/methods/delete.py
```diff
@@ -21,6 +21,27 @@ def test_unknown_resource(self):
         _, status = self.delete(url)
         self.assert404(status)
 
+    def test_bulk_delete_id_field(self):
```
nicolaiarocci: the new DELETE behavior should be explained in the docs, probably in the [Sub Resources](http://python-eve.org/features.html#sub-resources) of the features page.

Amedeo91:
Hi Nicola,

thanks for your review! :) I did the update you suggested to me.

Regards,
Amedeo

nicolaiarocci:
Could you please squash all commits into one?

Amedeo91:
Ciao Nicola,

commit uniti! :) 

Ciao,
Amedeo

Amedeo91:
Hi @nicolaiarocci, is there anything else I should do for this PR?

Regards,
Amedeo

nicolaiarocci:
Merged after a rebase and some fixes, see 63d87fc110434b9399f133e331dd69d89da2489d). 

Thanks!

PR #1032, opened 2017-06-21
Emdebbed perfomace

Amedeo91:
The purpose of this PR, is to provide some point of customization to the Framework consumer in order to put into place particular behavior as described per #1010.

This way, the framework will be more flexible and easier to put several use case into place.


Amedeo91:
I have some problem here and there with the commit history. I will do a new PR

PR #1038, opened 2017-07-18
Fix #1036 : allow to define mixed projections

### LINKED ISSUE #1036: Cannot define default projection and request specific field
Hi,

I am trying to exclude a field from a specific endpoint when requesting `/endpoint`. I use 
```
'datasource': {
            'projection': {
                'my_field_to_exclude': 0,
            }
        },
```
which works well. I can get the field on demand via `/endpoint?projection={"my_field_to_exclude":1}` (similar to http://python-eve.org/features.html#projection-filestorage).

However I want to be able to query the endpoint for a given field with `/endpoint?projection={"another_field":1}`. I expect to get a response with `another_field` only but I only get 
```
<link rel="collection" href="endpoint" title="mycollection"/>
<link rel="parent" href="/" title="home"/>
<_created>1970-01-01T00:00:00.000</_created>
<_etag>ce526180ec8bf46972f44049d7db1eaf25cca381</_etag>
<_id>731730</_id>
<_updated>1970-01-01T00:00:00.000</_updated>
```

If I remove the datasource projection, the field appears.

Any ideas ? I am missing anything ?

Thx !

Using Eve-0.7.4.

Hugoch:
The purpose of this PR is to fix the #1036 issue by allowing mixed projections. Mixed projections can be defined as default projection so that fields can be hidden by default and still requested in projections.
The use case is to reduce the amount of data sent by the API by hiding less used/data-extensive fields.

### DIFF
--- a/eve/io/base.py
+++ b/eve/io/base.py
@@ -439,6 +439,11 @@ class DataLayer(object):
                 # there's no standard projection so we assume we are in a
                 # allow_unknown = True
                 fields = client_projection
+        elif projection_ is not None:
+            # in case of mixed projection, only keep the exclusions
+            projection_vals = projection_.values()
+            if 0 in projection_vals and 1 in projection_vals:
+                fields = dict((k, v) for k, v in projection_.items() if not v)
 
         # If the current HTTP method is in `public_methods` or
         # `public_item_methods`, skip the `auth_field` check
--- a/AUTHORS
+++ b/AUTHORS
@@ -156,3 +156,4 @@ Patches and Contributions
 - dccrazyboy
 - mmizotin
 - xgdgsc
+- Hugo Larcher


eve/io/base.py
```diff
@@ -439,6 +439,11 @@ def _datasource_ex(self, resource, query=None, client_projection=None,
                 # there's no standard projection so we assume we are in a
                 # allow_unknown = True
                 fields = client_projection
+        elif projection_ is not None:
```
Amedeo91: elif projection_:

Amedeo91:
I do not see why you should put both inclusion and exclusion and why it should solve the issue you linked.
Can you please provide an example in order to make it clear?

Hugoch:
Given the following datasource for /endpoint
```
'datasource': {
            'projection': {
                'specialField': 0,
                'field1': 1
                ...
                'fieldn': 1
            }
        }
```
I can disable the `specialField`field in default view (embedded field with lots of data in my use case), and I get all the other field shown.
I can request the `specialField`field using a projection `/endpoint?projection={"specialField":1}`
... and I also can request a subset of the other fields `/endpoint?projection={"field1":1}` -> this is currently not possible without this PR.

Does that make any sense to you?

Amedeo91:
do you confirm that in both cases projection={"specialField":1}, /endpoint?projection={"field1":1} only the field at one will be returned? Otherwise, you should add a set-up to enable it. By the way: could you document this behavior aswell?

nicolaiarocci:
Hello, thanks for your contribution. If I get it right, this would essentially influence `datasource` projections, allowing for more selective picks on which fields are included and/or excluded whereas right now you can only exclude certain fields (and conseguently include all the others), or you can include some fields (and exclude others). Am I correct?

This would need test coverage as well as documentation upgrade.

huandzh:
hi, here is a PR for this issue at #1128 with tests and docs updated. found codes around projection setting quite fragile...

PR #1047, opened 2017-08-18
Support partial request for media resource

MarschHuynh:
I am creating an API for record and playback audio. I use Eve as a main lib for my project.
Without partial request, my player in my phone (iPhone) can not work.

So I change a little code to make Eve support partial request for the media resource.

Hope that makes sense.

Thanks.


Amedeo91:
Hi, thanks for your PR. Could you please provide test coverage and update the doc with this new behavior?

Regards,
Amedeo

PR #1050, opened 2017-08-20
Support partial request for media resource

MarschHuynh:
I am creating an API for record and playback audio. I use Eve as a main lib for my project.
Without partial request, my player in my phone (iPhone) can not work.

So I change a little code to make Eve support partial request for the media resource.

Hope that makes sense.

Thanks.

### DIFF
--- a/eve/endpoints.py
+++ b/eve/endpoints.py
@@ -11,6 +11,8 @@
     :copyright: (c) 2017 by Nicola Iarocci.
     :license: BSD, see LICENSE for more details.
 """
+import re
+
 from bson import tz_util
 from flask import abort, request, current_app as app, Response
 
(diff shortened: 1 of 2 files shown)


Amedeo91:
@nicolaiarocci: what do you think? Personally I do not see test coverage...

nicolaiarocci:
Hello and thanks for this. It is indeed a very cool feature. However, it would also need test coverage.

eve/endpoints.py
```diff
@@ -174,28 +176,76 @@ def media_endpoint(_id):
...
+        size = file_.length
+        byte1, byte2 = 0, None
+
+        m = re.search('(\d+)-(\d*)', range_header)
+        g = m.groups()
+
+        if g[0]:
+            byte1 = int(g[0])
+        if g[1]:
+            byte2 = int(g[1])
```
peterdemin: `g[1]` might be empty string, which will break `int('')`:
```
>>> int('')
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: ''
```

MarschHuynh: @nicolaiarocci Yes, I will update the test as soon as possible.

@peterdemin Thank you for correcting me.

nicolaiarocci:
This is still waiting on proper test coverage. Are you still planning to add it?

nicolaiarocci:
Thanks for the commit update. I really want to merge this but I can't in its current state. Still needs documentantion and test coverage.

nicolaiarocci:
Refactored, rebased, test-covered, and finally merged. See https://github.com/pyeve/eve/commit/9ee2b5baa97c361761ec73d7fa937ef82e9c310c

PR #1052, opened 2017-08-26
Update setup.py

lucifuturecoder:


nicolaiarocci:
Hello, thanks for this. However, I cannot accept this PR because we don't actively support python 3.1 and 3.2, which basically means that we don't run tests againsts those. Eve of course might still work on v3.1 and v3.2.

PR #1058, opened 2017-09-03
Handling lists as part of aggregation stage with parameters

### LINKED ISSUE #1025: Aggregation query parameter does not replace keys in the lists
I was working on an aggregation endpoint:

```python
things_recommended = {
    'url': 'things/recommended/',
    'datasource': {
        'source': 'things',
        'aggregation': {
            'pipeline': [
                {"$match": {"$or": [{"family": "$family_id"}, {"is_shared": True}]}},
                {"$addFields": {
                    "recommended_for_me":
                    {"$setIsSubset":[["$user_id"], "$recommended_to"]}
                }},
                {"$sort": SON([("recommended_for_me", -1), ("happened", -1)])}
            ]
        }
    }
}
```

However, I found that `$family_id` and `$user_id` are not getting replaced since the code only covers the case of a pure dictionary without lists.

See: https://github.com/pyeve/eve/blob/22ea4bfebc8b633251cd06837893ff699bd07a00/eve/methods/get.py#L126

I'm new to Eve, so I'm not sure that what I found worth an error report. Though it took me quite a time to debug.

Best regards,
Youngjae

PS, Eve is AWESOME.

idserge7:
This pull request resolves #1025


idserge7:
English is not my first language but I can try :)

So, there is a part called `MongoDB Aggregation Framework` but there is nothing about templates are only for `dicts`. Or should I provide an example for `list` with parameters?

Amedeo91:
Well, maybe just add a note where you underline that the api consumer can specify parameter either inside a dictionary or inside a list.

idserge7:
Ok. I'll think how to put it in a simple way.

nicolaiarocci:
Rebased and merged (https://github.com/pyeve/eve/commit/8b0515bb0262c7a14be296fef0a30f0f01f14dfb). Thanks!

PR #1070, opened 2017-10-10
use OrderedDict from backport_collections

carlwgeorge:
It is unnecessary to require the ordereddict module because the backport_collections module already has the OrderedDict class.  This change removes the ordereddict dependency, and also ensures that backport_collections is only required when needed.


nicolaiarocci:
Thanks!

PR #1091, opened 2017-12-03
Fix sanitization of nested queries.

Moschn:
A query was not fully traversed in the sanitization. Therefore the blacklist for mongo wueries could be bypassed, allowing for dangerous "$where" queries.


nicolaiarocci:
Merged and released with v0.7.5

Thanks!

PR #1092, opened 2017-12-15
Renderer classes

mpuhacz:
Hello,

Recently I was working on API which handles a lot of float values in responses. One of the issues I encountered was having NaN values unquoted in the JSON response. `simplejson` allows changing those NaNs to null values in the fly using the `ignore_nan=True` argument. Eve has already this nice feature that allows using custom `app.data.json_encoder_class` but even if the custom JSON encoder class has ignore_nan set to `True`, that is still ignored and falls back to the default value which is `False`. 

My idea is to have new key in the settings called `RENDERERS`. This is a tuple containing module path to Renderer classes. This way developer is allowed to simply extend the default Eve's functionality to new renderers. For example - using [ujson](https://pypi.python.org/pypi/ujson), [pyYaml](http://pyyaml.org/wiki/PyYAML).

This change is rather big, so I would love to get your feedback on this. If you feel like it's a good idea, I will continue working on it, updating docs and extending docstrings.

Best,
Marcin

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -94,6 +94,7 @@ Patches and Contributions
 - Mandar Vaze
 - Manquer
 - Marc Abramowitz
+- Marcin Puhacz
 - Marcus Cobden
 - Marica Odagaki
 - Mario Kralj
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -453,7 +453,6 @@ uppercase.
 
 ``META``                            Allows to customize the meta field. Defaults
                                     to ``_meta``
-                                    to ``_meta``.
 
 ``INFO``                            String value to include an info section, with the
                                     given INFO name, at the Eve homepage (suggested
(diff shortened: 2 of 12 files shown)


nicolaiarocci:
Hello, this is an interesting work, I like where you are heading. One concern of mine is backward compatibility. People who have `XML = False`  in their production settings will end up with XML enabled again after the upgrade. 

Are you planning to address this? Or are you just counting on breakage warning in release notes instead?

mpuhacz:
Good point. Both ways have their pros and cons, but I would go for breakage warning (or even throwing `DeprecationWarning`) because of simplicity. On the other hand, we can support both of `JSON` and `XML` and then create `RENDERERS` on the fly, but that would mean more complexity(an e.g. new method that checks for usage of deprecated features). I would love to keep it as simple as possible. What is your preferred approach?

nicolaiarocci:
I was going to suggest we issued deprecation warnings, but I see you went ahead and added the implementation already, excellent.

nicolaiarocci:
Please, go ahead with documentation updates.

mpuhacz:
@nicolaiarocci The docs are updated - hopefully I didn't miss anything. :smiley:

nicolaiarocci:
Rebased and merged: https://github.com/pyeve/eve/commit/215746f3c558ab71ea29b1f05dbe998b3cb948bb

Thank you!

Issue #1074, opened 2017-10-17
OPLOG =True is not working currently,


cst4049:
eve/methods/common.py
in v0.8
    if not config.OPLOG \
            or op not in config.OPLOG_METHODS\
            or resource in config.URLS[resource]: 
        return
in v0.7 and preview
    if not config.OPLOG or op not in config.OPLOG_METHODS: 
        return
it works in v0.7,and not work in v0.8

Amedeo91:
Share your configuration please

cst4049:
# oplog 
OPLOG = True
OPLOG_ENDPOINT = 'oplog'
I found common.py is different between  branch v0.6 to branch v0.8

zhangtemplar:
I have the same problem. 

```
OPLOG = True
OPLOG_METHODS = ['GET', 'DELETE', 'POST', 'PATCH', 'PUT']
```
this configuration doesn't work either.

What is `resource in config.URLS[resource]` targeted for?

cst4049:
oplog might be not record the GET method,
and your can try config.URLS[resource] not contain the resouce,
 like this: 
resource = "people" and the url must be not contain people 
you can use url like "human" and "aaa" something others just avoid "people"
the point is the url can't contain the resource name

zhangtemplar:
@cst4049 I have commented out `resource in config.URLS[resource]` but still doesn't work.

josemifv:
Hi all, 

I am currently working with Eve 0.7.4 and I have the same issue. After setting OPLOG=TRUE and OPLOG_ENDPOINT='oplog', the endpoint is created, but it is not populated.

After checking the source code, I noticed that the commit [a7350d](https://github.com/pyeve/eve/commit/a7350d08f4fc2e5155cf62e8f618401df14cb482#diff-825050a972dfddbdacccc7c0a830719b) fixed the issue, but the commit [662924](https://github.com/pyeve/eve/commit/6629241adb3c3615a0214fb2aa3f6da3929feb44#diff-825050a972dfddbdacccc7c0a830719b) changed the condition to the above-mentioned `resource in config.URLs[resource]` and the oplog_push does nothing as the initial condition always is True.

I changed the condition as coded in [a7350d](https://github.com/pyeve/eve/commit/a7350d08f4fc2e5155cf62e8f618401df14cb482#diff-825050a972dfddbdacccc7c0a830719b) and the OPLOG is working again in 0.7.4.

I hope it helps you :)

[...]

PR #1095, opened 2017-12-24
Oplog skipped even if config.OPLOG=True

lexhung:
Check for resource name validity is being reversed.


nicolaiarocci:
Hello, is this supposed to fix #1074? Thanks!

nicolaiarocci:
Would be great if you could add a regression test here.

nicolaiarocci:
Added a regression test (or I should say: fixed a false positive test), rebased and merged. Thanks.

Issue #1123, opened 2018-03-16
VALIDATE_FILTERS setting doesn't work for filter on sub-document field


lmoretto:
Activating the `VALIDATE_FILTERS` setting causes request with filters on sub-documents' fields to fail.

**Steps to reproduce**
Suppose you have a domain resource with the following schema:
```python
'collection': {
    'schema': {
        'sub_documents': {
            'type': 'list',
            'schema': {
                'type': 'dict',
                'schema' : {
                    'foo': {'type': 'string'}
                    'bar': {'type': 'integer'}
                }
            }
        }
    }
}
```
Performing a request like:
`http://<api_server>/collection?where={"sub_documents.bar": 5}`
returns the following response:
```json
{
    "_status": "ERR",
    "_error": {
        "code": 400,
        "message": "filter on '%s' is invalid"
    }
}
```

This seems to be caused in `eve/utils.py` at lines 403-406:
```python
if config.VALIDATE_FILTERS:
    res_schema = config.DOMAIN[resource]['schema']
    if key not in res_schema:
        return "filter on '%s' is invalid"
```
In this scenario, `key` will be `sub_documents.bar` which is not a valid key in the `res_schema` dict (whereas `sub_documents` is).

nicolaiarocci:
Excellent. Looks like we're missing support for sub-document parsing here. Feel free to submit a PR if so inclined, I will gladly review it.

PR #1125, opened 2018-03-23
1123 sub document validate filters

lmoretto:
This PR fixes issue #1123 

Also, commit bae6757 fixes the behavior of `ALLOWED_FILTERS` (and the related resource-specific `allowed_fillters`) settings in the same scenario as the one defined in #1123, i.e. filtering on a sub-document field.

### DIFF
--- a/eve/utils.py
+++ b/eve/utils.py
@@ -386,8 +386,22 @@ def validate_filters(where, resource):
 
     def validate_filter(filter):
         for key, value in filter.items():
-            if '*' not in allowed and key not in allowed:
-                return "filter on '%s' not allowed" % key
+            if '*' not in allowed:
+                def recursive_check_allowed(filter_key, allowed_filters):
+                    # Filter key can be a plain key (e.g. "foo") or a dotted key (e.g. "dict.sub_dict.bar").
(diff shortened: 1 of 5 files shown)


eve/utils.py
```diff
@@ -386,8 +386,27 @@ def validate_filters(where, resource):
 
     def validate_filter(filter):
         for key, value in filter.items():
-            if '*' not in allowed and key not in allowed:
-                return "filter on '%s' not allowed" % key
+            if '*' not in allowed:
+                def recursive_check_allowed(filter_key, allowed_filters):
+                    # Filter key can be a plain key (e.g. "foo") or a dotted
```
nicolaiarocci: I would rather move this comment to the documentation.

lmoretto: Hi @nicolaiarocci , thanks for your review!
Just to be sure, did you mean ALLOWED_FILTERS? Also, do you want me to remove this comment from the code once it will be present in the documentation?

nicolaiarocci: Yes, remove it from the code. And yes that's what I meant.

lmoretto: Ok, I updated the PR with the requested changes ;)

PR #1130, opened 2018-03-30
Fix PUT behavior with User-Restricted Resource Access

lmoretto:
This PR fixes the behavior of PUT requests when User-Restricted Resource Access is enabled.

More in particular it ensures that, under every circumstance, users are unable to overwrite items owned by other users.

### DIFF
--- a/eve/io/base.py
+++ b/eve/io/base.py
@@ -474,13 +474,7 @@ class DataLayer(object):
                             self.app.data.get_value_from_query(
                                 query, auth_field) != request_auth_value:
                             desc = 'Incompatible User-Restricted Resource ' \
-                                   'request. Request was for "%s"="%s" but ' \
-                                   '`auth_field` requires "%s"="%s".' % (
-                                       auth_field,
-                                       self.app.data.get_value_from_query(
-                                           query, auth_field),
-                                       auth_field,
-                                       request_auth_value)
+                                   'request.'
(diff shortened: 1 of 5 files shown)


eve/io/base.py
```diff
@@ -446,35 +456,40 @@ def _datasource_ex(self, resource, query=None, client_projection=None,
...
-                              'Resource request. '
-                              'Request was for "%s"="%s" but `auth_field` '
-                              'requires "%s"="%s".' % (
-                                  auth_field,
-                                  self.app.data.get_value_from_query(
-                                      query, auth_field),
-                                  auth_field,
-                                  request_auth_value)
-                              )
+                            desc = 'Incompatible User-Restricted Resource ' \
```
nicolaiarocci: this might disclose a little too much info to the client. Yes, the desc field is only emitted in debug mode, but you never know someone might leave debug on by mistake.

lmoretto: The description string was already present, I didn't change it. I simply restructured this piece of code to fix a Flake8 line too long error. Do you want me to change the description anyway? What should I write?

nicolaiarocci: Yes, you are right, it was there already. Let us obfuscate it a little bit, maybe by simply returning the first part of the message (drop the "request was for but xx was excpected")

lmoretto: done ;)

Issue #712, opened 2015-09-14
Patch/Update behavior - No way to remove a sub field


lucabelluccini:
If I have a document and I apply some changes to the updates in the update callback on a sub object of it, some fields are not cleaned.

E.g.

Document: { 'field1': 'test', 'field2': { 'subfield1': 1, 'subfield2': 2 } }

Changes in the update item callback: { 'field2' : { 'subfield2': 22 } }

Result: { 'field1': 'test', 'field2': { 'subfield1': 1, 'subfield2': 22 } }

nicolaiarocci:
Yes this is the intended behavious.

 >Nested document updates are merged with the original contents, we don't overwrite the whole thing. See #519 for details.

[Source](https://github.com/nicolaiarocci/eve/blob/develop/eve/methods/patch.py#L253-L254)

You might want to use PUT instead.

Issue #519, opened 2014-11-27
About how Eve manage readonly fields and how improve them in PATCH operations


pfreixes:
Current implementation of patch operation gets out from the original document those `readonly` fields that they don't have default because of Eve uses the `dict.update` method directly.

Eve says nothing about that the default values are a must for a read only fields, therefore those values got from `readonly` fields should be keep it at each update.

Have a look to the next snippet, it tries to figure out which `readonly` fields are in the original document and keep them.

```
>>> original = {"foo" : {"code": 100}}
>>> updates = {"hello": "world"}
>>> schema = {"foo": {'type': 'dict', 'schema': { 'code': { 'type': 'integer', 'readonly': True}}}, "hello": {"type": "string"}}
>>> update_keeping_readonly_fields(original, updates, schema) 
>>> original
{'hello': 'world', 'foo': {'code': 100}}
```

The idea behind `update_keeping_readonly_fields` is just make a update as a original `dict.update`method but keeping the **readonly** fields.

Continuing with this subject, **readonly** fields with no default values, the scenario explained below should be consider to change how Eve is calculatiing **Etag** today. 

Readonly fields are used along with the other fields to get the etag values but these,  at last read only,  can't be updated for the user. The idea behind **Etag** is give to the user a way to check that it holds the last version of the document , otherwise apply some kind of mechanism at user side to merge last changes . Then the use of **readonly** fields without default values as a fields to get the Etag value doesn't make sense.

These proposed changes can be used to handle **readonly** fields without default values aside of the Eve without breaking the idiosyncrasy of Eve. It means that other processes can directly update these fields using other tools and other requirements.

I'm working with a branch to implement these modifications but first of all I would like to know what you think of this.

pfreixes:
HI @nicolaiarocci have you considered the `etag` attribute? I can code this new feature and make a pull request.

nicolaiarocci:
@pfreixes sorry I did not get back to you on your last question. I think it can be useful so feel free to submit a PR for `etag` field-level attribute.

pfreixes:
@nicolaiarocci etag feature was inserted by the commit d7e9bd02 therefore forget it

PR #1140, opened 2018-05-05
Make merging of nested documents optional (enabled by default)

NotSpecial:
Rationale
---------

In the fix for #519, merging of nested documents on `PATCH` was introduced,
essentially a `original.update(updates)` is performed.
So far, the only way to avoid this is to use `PUT` (see #712), however,
this is not always desirable.

We have a use case where we only want to update a nested field of a
document, therefore the appropriate HTML verb is `PATCH` since we do not want
to replace the whole document.
In this nested field, we need to remove a key, which, due to the merging,
is currently impossible.

Changes
-------

- Introduce a new setting, `MERGE_NESTED_DOCUMENTS`, to control this behaviour
  globally or per-resource.
  Per default, this is set to `True`, so that nothing changes compares to the
  current version of Eve except when the configuration is changed explicitly.
- Add a test
- Add documentation

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,6 +8,7 @@ Development Lead
 
 Patches and Contributions
 `````````````````````````
+- Alexander Dietmüller
 - Alexander Hendorf
 - Amedeo Bussi
 - Andreas Røssland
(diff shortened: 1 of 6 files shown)


nicolaiarocci:
Rebased and merged https://github.com/pyeve/eve/commit/72d8f8dbf824896a02250488c7215658374e434d

Thanks!

PR #1148, opened 2018-05-14
use simplejson module everywhere

### LINKED ISSUE #1072: json or simplejson
I've noticed the codebase has a mixed usage of the standard library json and the simplejson module.  Is this intentional?  Is there a preference?  If simplejson is preferred, I'd be happy to send a pull request to implement try/except blocks everywhere to prefer it with a standard library fallback.  Something like:

```python
try:
    import simplejson as json
except ImportError:
    import json
```

carlwgeorge:
resolves #1072

### DIFF
--- a/eve/methods/get.py
+++ b/eve/methods/get.py
@@ -13,7 +13,7 @@
 import math
 
 import copy
-import json
+import simplejson as json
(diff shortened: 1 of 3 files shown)


nicolaiarocci:
Thanks!

Issue #1175, opened 2018-07-25
AttributeError in deprecated_renderers_settings fonction


quentinpraz:
### Actual Behavior

During the initialization of the flask API, an error is raised in _deprecated_renderers_settings()_ function.
The error is the following:
```pytb
  File "/run.py", line 2, in <module>
    app = Eve()
  File "/usr/lib/python2.7/site-packages/eve/flaskapp.py", line 141, in __init__
    self.load_config()
  File "/usr/lib/python2.7/site-packages/eve/flaskapp.py", line 266, in load_config
    self.check_deprecated_features()
  File "/usr/lib/python2.7/site-packages/eve/flaskapp.py", line 292, in check_deprecated_features
    deprecated_renderers_settings()
  File "/usr/lib/python2.7/site-packages/eve/flaskapp.py", line 280, in deprecated_renderers_settings
    self.config['RENDERERS'] = default_settings.RENDERERS.copy()
AttributeError: 'list' object has no attribute 'copy'
```
This is probably du to the _copy()_ function that was introduced in python 3.3.
Solution would be to replace _default_settings.RENDERERS.copy()_ by _list(default_settings.RENDERERS) or default_settings.RENDERERS[:]_
### Environment

* Python version: 2.7.15
* Eve version: 0.8

PR #1177, opened 2018-07-25
Fixing issue 1175

quentinpraz:
This PR fixes issue 1175. According to Python documentation ([here](https://docs.python.org/3/library/stdtypes.html#mutable-sequence-types)), the copy() function is only available for Python 3.3+. To support Python 2.7.X, a change is needed.
According to the documentation, list.copy() is equal to list[:].


nicolaiarocci:
#1175 

Added a regression test, then merged. Thanks!

PR #1201, opened 2018-10-08
Support for resource .xml and .json endpoints

luiscoms:
If endpoint ends with .xml or .json, use its renderer


nicolaiarocci:
Thank you for this. However, in REST the de-facto way to ask for different rendering is the `Accept` header, which is supported already. An alternative/non-standard way to achieve the same results would add complexity with little value.

Issue #1186, opened 2018-09-02
Adding HATEOAS support for embedded fields


smeng9:
Hi,

I think it would be useful to add a reference sort of thing to the returned embedded field, if we want the user to navigate.

---

In https://eve-demo.herokuapp.com/ demo there are two domains -- works and people.
For now when we visit https://eve-demo.herokuapp.com/works, the embedded owner field in each of returned resource is just an objectid string with no href.

### Expected Behavior

visit https://eve-demo.herokuapp.com/works

```xml
<resource href="works/5b89b1b091a5d0000495f56a" title="Work">
<_created>Fri, 31 Aug 2018 21:22:56 GMT</_created>
<_etag>9699888bc6e1d48fb52d3b573f302a24112802e8</_etag>
<_id>5b89b1b091a5d0000495f56a</_id>
<_updated>Fri, 31 Aug 2018 21:22:56 GMT</_updated>
<description>Description #25</description>
<owner href="people/5b89b1b091a5d0000495f54d">5b89b1b091a5d0000495f54d</owner>
<title>Book Title #25</title>
</resource>
```

visit https://eve-demo.herokuapp.com/works?embedded={%22owner%22:1}

```xml
<resource href="works/5b89b1b091a5d0000495f56a" title="Work">
<_created>Fri, 31 Aug 2018 21:22:56 GMT</_created>
<_etag>9699888bc6e1d48fb52d3b573f302a24112802e8</_etag>
<_id>5b89b1b091a5d0000495f56a</_id>
<_updated>Fri, 31 Aug 2018 21:22:56 GMT</_updated>
<description>Description #25</description>
<owner href="people/5b89b1b091a5d0000495f54d">
  <_created>Fri, 31 Aug 2018 21:22:56 GMT</_created>
  <_etag>103dab458e6e1cf1a70b74f770bb0cac9075a1c0</_etag>
  <_id>5b89b1b091a5d0000495f54d</_id>
  <_updated>Fri, 31 Aug 2018 21:22:56 GMT</_updated>
  <born>Wed, 25 Feb 1987 17:00:00 GMT</born>
  <firstname>Serena</firstname>
  <lastname>Love</lastname>
  <location>
  <address>363 Brannan St</address>
  <city>San Francisco</city>
  </location>
  <role>author</role>
</owner>
<title>Book Title #25</title>
</resource>
```

### Actual Behavior

visit https://eve-demo.herokuapp.com/works

```xml
[...]

PR #1204, opened 2018-10-12
[WIP] Enhancement hateoas #1186

smeng9:
Support for xml renderer and test cases is still in progress


nicolaiarocci:
For reference, see #1186

PR #1210, opened 2018-11-22
Prune the stages from aggregation pipeline whose condition is not set…

zhangtemplar:
… or empty.


zhangtemplar:
@nicolaiarocci how should I mute the warnings for:
```
eve/tests/renders.py:201:22: W605 invalid escape sequence '\d'
eve/tests/renders.py:201:27: W605 invalid escape sequence '\.'
eve/tests/renders.py:201:35: W605 invalid escape sequence '\.'
```

zhangtemplar:
OK, it turns out to be much more complex than I thought.

I cannot simply prune out all values starting with $, as MongoDB some times strings starts with $ also refer to a field in the document.

The alternative is that, user should always provide parameter for the variables in the aggregation pipeline; however, for the variables which are intended to omit/ignore, set `{}` to those variables.

nicolaiarocci:
rebased and merged. Thanks.

zhangtemplar:
Cool

PR #1223, opened 2019-01-27
Feature new event hook for diffs

einarhuseby:
To plug the hole left from not having event hooks fire on diffs. A use case could be anonymizing documents via `on_fetched_*`. Accessing with `version=diffs` would expose the documents. 

This alternative introduces `on_fetched_diff`event hooks to allow wiring up  `version=diffs`

### DIFF
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -1270,6 +1270,12 @@ Let's see an overview of what events are available:
 |       |        |      +--------------------------------------------------+
 |       |        |      || ``on_fetched_item_<resource_name>``             |
 |       |        |      || ``def event(response)``                         |
+|       +--------+------+--------------------------------------------------+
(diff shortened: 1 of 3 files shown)


nicolaiarocci:
In order to not to break existing callbacks, I feel like this approach is more sensible than the one proposed with #1224. Thoughts?

einarhuseby:
Initially I favored #1224 over this since it really should be a catch all. But reconsidered due to the breakage. So I agree, go for this if you want to merge it.

einarhuseby:
Actually, firing for every partial document is probably not the most sensible thing to do. Might be better to emit all versions as one response, which will provide the callback with the `id_field`. The naming of the hook also implies the callback should expect an array of diffs

nicolaiarocci:
Sounds reasonable.

Issue #475, opened 2014-10-11
On fetched event hooks for viewing all versions or version diffs


cuibonobo:
There's currently a `TODO` listed in the `getitem` method that is preventing requests with `?version=all` or `?version=diffs` from executing `on_fetched` events. (See [here](https://github.com/nicolaiarocci/eve/blob/develop/eve/methods/get.py#L286).) The project I'm working on uses event hooks extensively to examine the content of a document before sending it, so I'm particularly interested what needs to be done to get this working.

For my project I just commented out the `if` statement that's preventing the event hooks from firing and everything seems to be working fine. What else is missing?

/cc @joshvillbrandt because I know document versioning is his baby.

PR #1224, opened 2019-01-27
Allow event fetched on diffs#475

einarhuseby:
See PR #1223 for the other alternative introducing `on_fetched_diffs`.

This PR allows `version=diffs` to run when `on_fetched*` event hooks are wired up.

This makes all item and resource fetches fire the callback. A use case could be anonymizing documents via on_fetched_*. Accessing with version=diffs would expose the documents.

As discussed in #475 and inline code notes `version=diffs` will return partial documents which can have unexpected results if it's not handled correctly in the callback.

### DIFF
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -1375,9 +1375,11 @@ the items as needed before they are returned to the client.
     >>> app.on_fetched_item_contacts += before_returning_contact
 
 It is important to note that fetch events will work with `Document
-Versioning`_ for specific document versions or accessing all document
-versions with ``?version=all``, but they *will not* work when accessing diffs
-of all versions with ``?version=diffs``.
+Versioning`_ for specific document versions like ``?version=5``, accessing all
+document versions with ``?version=all``, and accessing diffs of all versions
+with ``?version=diffs``. When working with versioning, care should be taken in
+the registered callback to handle possible schema differences or partial
+documents. Diffs by design is partial documents.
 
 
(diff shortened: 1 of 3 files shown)


einarhuseby:
Merged the changes as discussed in #1223 to this PR instead of making another PR since this is labelled for 0.8.2.

The callback now gets the entire list of diffs, which I think makes sense. Also updated the tests in TestCompleteVersioning; `test_on_fetched_diffs` and `test_on_fetched_diffs_contacts`.

Hateos _links seems to get an extra `&version=diffs` but I haven't looked further into that. Not sure if it's caused by this PR.

nicolaiarocci:
> Hateos _links seems to get an extra &version=diffs but I haven't looked further into that. Not sure if it's caused by this PR.

Can you please investigate further on this, and report back. Thanks.

einarhuseby:
Looks like `default_params` is missing version https://github.com/pyeve/eve/blob/7a8a6c269cd4b05f8afbf5f747f62fe31dd66964/eve/methods/get.py#L622-L634 and then it gets added as `other_params` in https://github.com/pyeve/eve/blob/7a8a6c269cd4b05f8afbf5f747f62fe31dd66964/eve/utils.py#L296

Maybe `QUERY_VERSION` instead of `VERSION_PARAM` would make sense?

nicolaiarocci:
rebased and merged, thank you

PR #1225, opened 2019-01-27
Fixes #1069 following data_relation.field

### LINKED ISSUE #1069: Embedding only follows _id
I have the following schemes:

    nodes = {
        'type': 'dict',
        'schema': {
            'hardware': {
                'type': 'string',
                'data_relation': {
                    'resource': 'nodes_hw',
                    'field': 'name',
                    'embeddable': True
                }
            }
        }
    }

    nodes_hw = {
        'type': 'dict',
        'schema': {
            'name': {
                'type': 'string',
                'required': True,
                'unique': True
            }
        }
    }

When now querying 

    /nodes?embedded={"hardware":1}

the resulting value for hardware is null.

The following lines seem to be the problem
https://github.com/pyeve/eve/blob/cee2f7ec13976f5bc3910f7058648a8b33ec882c/eve/methods/common.py#L826-L829

Internally, in nodes_hw a lookup is performed with _id = VALUE_OF_HARDWARE.

Changing this line to use data_relation['field'] yields the expected result.
Then the internal lookup is correctly performed with name = VALUE_OF_HARDWARE
As result, hardware contains a dictionary with the embedded nodes_hw object.

Am I doing something wrong, is this not wanted or is this a bug?

einarhuseby:
Embedded resources, `data_relation["field"]` should be honored.

See #1069

### DIFF
--- a/eve/methods/common.py
+++ b/eve/methods/common.py
@@ -984,7 +984,8 @@ def generate_query_and_sorting_criteria(data_relation, references):
         id_field_name = (
             "_id"
             if isinstance(reference, DBRef)
-            else config.DOMAIN[subresource]["id_field"]
+            else data_relation.get("field", False)
+            or config.DOMAIN[subresource]["id_field"]
         )
         id_field_value = reference.id if isinstance(reference, DBRef) else reference
         query["$or"].append({id_field_name: id_field_value})


nicolaiarocci:
rebased and merged, thanks.

Issue #1202, opened 2018-10-11
DeprecationWarning on newer pymongo


nicolaiarocci:
`DeprecationWarning: count is deprecated. Use Collection.count_documents instead.` is returned by pymongo 3.7.2. Consider updating the code to silence the warning, and of course pin to pymongo 3.7+

einarhuseby:
@nicolaiarocci - I was looking into this and soon found that
```
cursor = app.data.find(resource, req, lookup)
count = cursor.count()
```
needs to be replaced like
```
count = collection.count_documents({"foo": "bar"})
```
For Eve I think either a new `count` method for the datalayer `count = app.data.count(resource, req, lookup)` or handle it in  `find` and change the return signature from `cursor` to `count, cursor`.

Both will break the current base datalayer. Personally I favor the latter. What are your thoughts on this?

Also pymongo 3.7 added [estimated_document_count]( http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.estimated_document_count). See also this [SO answer](https://stackoverflow.com/questions/52236594/why-is-pymongo-count-documents-is-slower-than-count)

This could possible be of use;  `OPTIMIZE_PAGINATION_FOR_SPEED` could be replaced with something like `PAGINATION_STRATEGY` and a corresponding resource override `pagination_strategy` with options full count, estimated count and no count.

I could provide a PR to see how it pans out.

nicolaiarocci:
> I could provide a PR to see how it pans out.

Please do.

PR #1226, opened 2019-01-29
Deprecation warning newer pymongo #1202

einarhuseby:
PR to get some comments on the implementation ref discussion for #1202.  No tests yet.

Main takeaways are:

- needs pymongo>=3.7
- breaks datalayer `find()` return signature from `cursor` to `count, cursor`
- introduces new global config setting `PAGINATION_STRATEGY` and per resource `pagination_strategy`

 `PAGINATION_STRATEGY` will probably make `OPTIMIZE_PAGINATION_FOR_SPEED` obsolete. Options are

- `None` no count
- `full` pymongos `count_documents` which do an aggregation query over all documents hence is accurate, equivalent to deprecated `cursor.count()`
- `estimated` pymongos `estimated_document_count` which estimates the number of documents based on collection statistics

I haven't tried  `count_documents` hint and max execution time options, could possibly be something to investigate?

I did some simple benchmarks with wrk. On the smaller collections there is not much difference, but it's appearent that `count_documents` becomes increasingly slower with collection size.

On this PR,  `PAGINATION_STRATEGY`:

- `None` 216 requests in 20.02s
- `estimated` 191 requests in 20.03s
- `full` 9 requests in 20.04s

On pyeve/master,  `OPTIMIZE_PAGINATION_FOR_SPEED`:

- `False` 9 requests in 20.03s
- `True` 206 requests in 20.03s

Tests done on an indexed collection just shy of 5m documents, one connection, to be taken as an indication only.

I did observe `estimated` to be accurate on all requests. Also, the very first database hit for `estimated` takes around 1/3 of the time of an `full` then drops by a factor of 10 and is constant for all following requests. I assume that `estimated` will be accurate and almost as fast as no counting for large collections with few updates. With frequent updates it would probably become slower but still a lot faster than doing the `full` count. I don't know about accuracy on frequent updates.

I think this could be of great use especially for those with large collections.

Feedback appreciated!

### DIFF
--- a/eve/default_settings.py
+++ b/eve/default_settings.py
@@ -11,6 +11,8 @@
     :copyright: (c) 2017 by Nicola Iarocci.
     :license: BSD, see LICENSE for more details.
 
+    .. versionchanged:: 0.8.2
+        'PAGINATION_STRATEGY' options are None, full and estimate and set to full.
     .. versionchanged:: 0.8
         'RENDERERS' added with XML and JSON renderers.
         'JSON' removed.
@@ -164,6 +166,9 @@ PROJECTION = True  # projection enabled by default
 PAGINATION = True  # pagination enabled by default.
 PAGINATION_LIMIT = 50
 PAGINATION_DEFAULT = 25
+HEADER_TOTAL_COUNT = "X-Total-Count"
+PAGINATION_STRATEGY = "full"
+OPTIMIZE_PAGINATION_FOR_SPEED = False
 VERSIONING = False  # turn document versioning on or off.
 VERSIONS = "_versions"  # suffix for parallel collection w/old versions
 VERSION_PARAM = "version"  # URL param for specific version of a document.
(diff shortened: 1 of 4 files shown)


einarhuseby:
Something is not right with pagination, and after some RTFM it's clear to me that `estimated_document_count` *only* gets the total number of documents - and not by filter.

However,  `estimated_document_count` should still be useful when no filter/counting all documents in a collection.

Also need to look into the replacement of query operators for [count_documents](http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.count_documents)

nicolaiarocci:
Hi, what is the state on this PR, still WIP?

I like the direction you are going.

nicolaiarocci:
Alright, I took a slightly different approach (see https://github.com/pyeve/eve/commit/15f5e650724b2a7503312f739f5442e5e2269e20), thank you for showing me the path.

Also, I focused on the deprecation warning. Feel free to open a new PR with work on pagination optimization.

nicolaiarocci:
Amended to only run `count_documents()` when required, see https://github.com/pyeve/eve/commit/ac3bd78a9dfcad38e2fe478c0f35f66ac290c37c.

PR #1228, opened 2019-01-31
Added VERSION_PARAM to default_params

einarhuseby:
See discussion in #1224, VERSION_PARAM missing and ends up being added to other params resulting in a double `&version=diffs&version=diffs` in `_links`


nicolaiarocci:
rebased and merged. thank you.

Issue #920, opened 2016-10-13
Within-one-second-precision PATCH should not raise 412


moreiramarti:
Hello, 

Thanks for you API it's really cool.
I have a small question about patch requests.

I disabled conditional requests with IF-MATCH=False but when I try to send 2 patch requests within the same second, I get: 
"code": 412, "message": "Client and server etags don't match"

Maybe I'm not using it right.

Can you help me with this ?

Thanks,
Martinho MOREIRA

marceljanerfont:
I have got the same problem. Digging in the source code I see in methods/patch.py line 174:

```
# some datetime precision magic
updates[config.LAST_UPDATED] = \
                datetime.utcnow().replace(microsecond=0)
```

They remove microseconds resolution, and if the PATCH is neutral (document is not changed) and '_updated' is not modified because you are in the same second window, therefore it returns 412, because  these lines in io/mongo/mongo.py:

```
result = coll.replace_one(filter_, changes) if replace else \
                coll.update_one(filter_, changes)
if result and result.acknowledged and result.modified_count == 0:
                raise self.OriginalChangedError()
```

A quick fix would be adding a changing field like timestamp or random value in your schema and modify it in each PATCH.

@nicolaiarocci Why do you discard microsecond precision at '_update' field?

nicolaiarocci:
@marceljanerfont  `_update` follows the RFC1123 format which has a 1 second precision.

This looks like a bug. If `IF_MATCH` is disabled it should not return a 412, even when the second update is neutral. I'm working on a fix.

elecay:
One thing that keeps bother me is that if I make two PATCHs in the same second without changing the document, then I will receive a 412 (this is, a **OriginalChangedError**); don't really understand why this _result.modified_count == 0_ should raise an exception. Maybe you consider that if there is no changes then the _filter__ is wrong?

nicolaiarocci:
Hi folks, please see if the current patch solves your issues. I got it wrong the first time around, but hopefully it's fixed now.

elecay:
Awesome Nicola! Thanks for your work and the quick response! Regards.

marceljanerfont:
It looks fine Nicola, thanks for this awesome tool.

nicolaiarocci:
Hi everyone. Would any of you take a look at #1231 and report their thoughts on it? I'm willing to merge it soon. Thanks.

PR #1232, opened 2019-03-03
Atomic Concurrency Checks

ceturc:
Atomic Concurrency Checks

See pyeve/eve#1231 for summary

Atomic Concurrency Checks

Prior to commit 54fd697 from 2016-November, ETags would be verified
twice during a patch. One ETag check would be non-atomic by Eve,
then again atomically by MongoDB during app.data.update(filter).
The atomic ETag check was removed during issue #920 in 54fd697

When running Eve in a scale-out environment (multiple processes),
concurrent simultaneous updates are sometimes allowed, because
the Python-only ETag check is not atomic.

There is a critical section in patch_internal() between get_document()
and app.data.update() where a competing Eve process can change the
document and ETag.

The included test simulates another process changing data & ETag
during the critical section. The test patches get_document() to return
an intentionally wrong ETag.

It is unclear if this patch causes any regression issues with #920

### DIFF
--- a/eve/io/mongo/mongo.py
+++ b/eve/io/mongo/mongo.py
@@ -469,9 +469,18 @@ class Mongo(DataLayer):
 
         coll = self.get_collection_with_write_concern(datasource, resource)
         try:
-            coll.replace_one(filter_, changes) if replace else coll.update_one(
-                filter_, changes
+            result = (
+                coll.replace_one(filter_, changes)
+                if replace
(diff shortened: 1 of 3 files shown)


nicolaiarocci:
Nice, thanks. Willing to merge. I'll wait from feedback from the folks involved with #920.

ceturc:
Thank you very much, @nicolaiarocci

PR #1238, opened 2019-03-22
To be compatible with werkzeug version 0.15.1

ph0nemyintkyaw:
In werkzeug version 0.15.1, response automatically include WWW-Authenticate header and they don't accept response argument anymore which is crashing when you add Auth in your domain and it is being unauthenticated.

### DIFF
--- a/eve/auth.py
+++ b/eve/auth.py
@@ -146,10 +146,7 @@ class BasicAuth(object):
         """ Returns a standard a 401 response that enables basic auth.
         Override if you want to change the response and/or the realm.
         """
-        resp = Response(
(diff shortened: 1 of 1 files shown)


nicolaiarocci:
Hello thanks for this. I pinned v0.14.1 just yesterday because of this. However 1 test is still failing, would you push a fix for it as well?

ph0nemyintkyaw:
Hi, I'm also planning to work on it in this weekend. I'll try.

nicolaiarocci:
Hey there, thanks for the attempt. It turned out that this was more subtle than expected, so I pushed a fix myself, see https://github.com/pyeve/eve/commit/63ddfa8181cc5b108b0ece154e0177989dbc519b.

PR #1253, opened 2019-04-03
PATCH not working as expected for nested document

zhangtemplar:


zhangtemplar:
Why it keep complaining on the lint issues while my computer is fine.

Issue #1234, opened 2019-03-12
PATCH incorrectly normalizes default values in sub-documents


zhangtemplar:
Consider the following schema:

```
'entity': {
  'name': {
    'type': 'string',
    'required': True },
  'contact': {
    'type': 'dict',
    'required': True,
    'schema': {
      'phone': {
        'type': 'string',
        'required': False,
        'default': '1234567890' },
      'email': {
        'type': 'string',
        'required': False,
        'default': 'abc@efg.com' },
    }
  }
}
```

Assume I have data like:
```
{
  "name": "test",
  "contact: {
    "phone": "0987654321",
    "email": "uvw@xyz.com"
  }
}
```

When I do a patch with data `{"contact.phone": "1111111111"}` to update the phone only:

### Expected behavior

The data will be changed to 
```
{
  "name": "test",
  "contact: {
    "phone": "1111111111",
    "email": "uvw@xyz.com"
  }
}
```

### Actual Behavior

But the data I get is:
```
{
  "name": "test",
  "contact: {
    "phone": "1111111111",
    "email": "abc@efg.com"
  }
}
```

Yes, the email is also changed: since email is not provided in PATCH body, its default value is used. Is it the intended behavior for eve?

### Environment

* Python version: 3.5.2
* Eve version: 0.8.1

zhangtemplar:
To resolve this problem, I think the solution would be changing the `eve/validation.py`'s `validate_update` method and set `normalize` to be `True`.

```
    def validate_update(self, document, document_id, persisted_document=None):
        """ Validate method to be invoked when performing an update, not an
        insert.

        :param document: the document to be validated.
        :param document_id: the unique id of the document.
        :param persisted_document: the persisted document to be updated.
        """
        self.document_id = document_id
        self.persisted_document = persisted_document
        return super(Validator, self).validate(document, update=True, normalize=False)
```

zhangtemplar:
@nicolaiarocci D1235 is ready.

nicolaiarocci:
@zhangtemplar see my comment in the PR

nicolaiarocci:
I reverted 9b2ba0f while we wait for a non-failing patch.

PR #1254, opened 2019-04-03
PATCH incorrectly normalizes default values in sub-documents

zhangtemplar:
For issue #1234


nicolaiarocci:
Thanks for this. However, this approach could not solve the issue. The added test was good, however, so I rebased the PR, added an alternative solution, then merged. 

See https://github.com/pyeve/eve/commit/52157710dbce300025f0fd3c86139c01a7c80df2 for the complete solution.

PR #1256, opened 2019-04-04
Key-value pair order in DBRef are honored when encoding

### LINKED ISSUE #1255: Insertion failure when replacing unknown field with dbref value
### Background
When javascript processes JSON or when MongoDB handles DBRef they keep the key value pairs ordered.
As we can see from the MongoDB source code, the expected format for DBRef should always have $ref in the first place, $id in the middle place, and $db in the last place as {$ref:, $id:, $db:}
https://github.com/mongodb/mongo/blob/b21647a5e19e5980186e57f77be0b1e0ef92a087/src/mongo/db/update/storage_validation.cpp#L86

### Expected Behavior

Tell us what should happen.

Below are the steps to reproduce the issue
1. define a ``works'' resource with ``allow_unknown'' set to True, and ``schema'' set to {}
2. create an item with {"title": "my work", "owner": {"$ref": "people", "$id": "objectid_in_people"}}
curl 'http://127.0.0.1:5000/works/ -X POST-H 'content-type: application/json' -d '{"title": "my work", "owner": **{"$ref": "people", "$id": "objectid_in_people"}**}'
3. get the created item using curl http://127.0.0.1:5000/works/objectid_in_work and we will get
{"title": "my work", "owner": **{"$id": "objectid_in_people", "$ref": "people"}}**
4. Try to PUT the data from step 3 to the same item endpoint

We noticed the order of $id and $ref are swapped in the GET response.
If we swap the order of $id and $ref in owner field again to {"$ref": "people", "$id": "objectid_in_people"}, then we are able to successfully replace the document.

When we are returning a dictionary for the DBRef, we are not using the OrderedDict, and we lose the order. 
https://github.com/pyeve/eve/blob/48d9c25cdca04c2f5a31ce01a1310d493fc67eda/eve/io/mongo/mongo.py#L59

### Actual Behavior

Tell us what happens instead.

```pytb
File "/usr/local/lib/python3.7/site-packages/eve/io/mongo/mongo.py", line 509, in change_request
    coll.replace_one(filter, changes) if replace else coll.update_one( 
File "/usr/local/lib/python3.7/site-packages/pymongo/collection.py", line 925, in replace_one collation=collation, session=session), 
File "/usr/local/lib/python3.7/site-packages/pymongo/collection.py", line 851, in _update_retryable _update, session) 
File "/usr/local/lib/python3.7/site-packages/pymongo/mongo_client.py", line 1248, in _retryable_write return self._retry_with_session(retryable, func, s, None) 
File "/usr/local/lib/python3.7/site-packages/pymongo/mongo_client.py", line 1201, in _retry_with_session return func(session, sock_info, retryable) 
File "/usr/local/lib/python3.7/site-packages/pymongo/collection.py", line 847, in _update retryable_write=retryable_write) 
File "/usr/local/lib/python3.7/site-packages/pymongo/collection.py", line 818, in _update _check_write_command_response(result) 
File "/usr/local/lib/python3.7/site-packages/pymongo/helpers.py", line 217, in _check_write_command_response _raise_last_write_error(write_errors) 
File "/usr/local/lib/python3.7/site-packages/pymongo/helpers.py", line 199, in _raise_last_write_error raise WriteError(error.get("errmsg"), error.get("code"), error) 
pymongo.errors.WriteError: Found $id field without a $ref before it, which is invalid.
```

### Environment

* Python version: 3.7.3
* Eve version: 0.8.1

### Suggested Fix
We can switch to OrderedDict to preserve the order. When user post DBRef data structures in JSON, they should also be aware of the order of the key value pairs.

smeng9:
Closes #1255 

Follow-ups:
In the POST or PUT request to a resource that allows unknown field, if the key-value pair of DBRef does not follow the order defined in MongoDB, we need to return a correct HTTP code instead of giving a 500.

### DIFF
--- a/eve/io/mongo/mongo.py
+++ b/eve/io/mongo/mongo.py
@@ -24,6 +24,7 @@ from pymongo import WriteConcern
 from werkzeug.exceptions import HTTPException
 import decimal
 from bson import decimal128
+from collections import OrderedDict
 
 from eve.auth import resource_auth
 from eve.io.base import DataLayer, ConnectionException, BaseJSONEncoder
@@ -41,6 +42,9 @@ class MongoJSONEncoder(BaseJSONEncoder):
     """ Proprietary JSONEconder subclass used by the json render function.
     This is needed to address the encoding of special values.
 
+    .. versionchanged:: 0.8.2
+       Key-value pair order in DBRef are honored when encoding. Closes #1255.
+
     .. versionchanged:: 0.6.2
        Do not attempt to serialize callables. Closes #790.
 
@@ -57,7 +61,9 @@ class MongoJSONEncoder(BaseJSONEncoder):
             # (and we probably don't want it to be exposed anyway). See #790.
             return "<callable>"
         if isinstance(obj, DBRef):
-            retval = {"$id": str(obj.id), "$ref": obj.collection}
+            retval = OrderedDict()
+            retval["$ref"] = obj.collection
+            retval["$id"] = str(obj.id)
             if obj.database:
                 retval["$db"] = obj.database
             return retval


nicolaiarocci:
rebased and merged. thanks! Go ahead with the PUT/POST fixes.

PR #1257, opened 2019-04-06
Return 400 if update/replace with malformed DBRef field.

smeng9:
Closes follow up of #1256 

Additional fix for #1255
This time even if ``json_sort_keys'' is set to True, the key-value pairs in DBRef will keep their order.


nicolaiarocci:
rebased and merged. thanks.

PR #1261, opened 2019-04-22
Add field `normalize_document_for_patch` to control whether apply normalize for validator of patch request.

zhangtemplar:
By default it is true, in this case the fields which are not included in the patch body will be reset to the default value specified in the schema.

@nicolaiarocci sorry for multiple requests for a single issue. My previous request will break existing tests and this requests pass test in my repository.

#1234

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -776,6 +776,11 @@ uppercase.
                                     note that with the default Mongo layer,
                                     setting this to ``False`` will result in an
                                     error. Defaults to ``True``.
+``NORMALIZE_DOCUMENT_FOR_PATCH``    If ``True``, the patch document will be
+                                    normalized according to schema. This means
+                                    if a field is not included in the patch
+                                    body, it will be reset to the default value
+                                    in its schema.
 
 =================================== =========================================
 
@@ -1120,6 +1125,11 @@ always lowercase.
                                 with the default Mongo layer, setting this to
                                 ``False`` will result in an error. Defaults to
                                 ``True``.
+``normalize_document_for_patch``    If ``True``, the patch document will be
+                                    normalized according to schema. This means
+                                    if a field is not included in the patch
+                                    body, it will be reset to the default value
+                                    in its schema.
 
 =============================== ===============================================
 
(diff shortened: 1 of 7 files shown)


zhangtemplar:
@nicolaiarocci a friendly reminder

docs/config.rst
```diff
@@ -776,6 +776,11 @@ uppercase.
                                     note that with the default Mongo layer,
                                     setting this to ``False`` will result in an
                                     error. Defaults to ``True``.
+``NORMALIZE_DOCUMENT_FOR_PATCH``    If ``True``, the patch document will be
+                                    normalized according to schema. This means
+                                    if a field is not included in the patch
+                                    body, it will be reset to the default value
+                                    in its schema.
```
nicolaiarocci: Two things here:
1) I would add an explanation of what happens when set to False (a field with a default schema value will be ginored if absent aka, won't be reset);
2)  I would add "Defaults to True at the end of the paragraph.

docs/config.rst
```diff
@@ -1120,6 +1125,11 @@ always lowercase.
                                 with the default Mongo layer, setting this to
                                 ``False`` will result in an error. Defaults to
                                 ``True``.
+``normalize_document_for_patch``    If ``True``, the patch document will be
+                                    normalized according to schema. This means
+                                    if a field is not included in the patch
+                                    body, it will be reset to the default value
+                                    in its schema.
```
nicolaiarocci: same as above

docs/features.rst
```diff
@@ -2319,6 +2319,37 @@ The stage ``{"$match": { "name": "$name", "time": "$time"}}`` in the pipeline wi
...
+        'required': False,
+        'default': 'abc@efg.com' },
+    }
+  }
+}
+```
+
+Two notations ``contact: { email: 'an email'}`` and ``contact.email: 'an email'`` can be used to update the `email` field embedded in `contact` field.
+
+
```
nicolaiarocci: Since we're adding a special noe on PATCH, I feel that this is the right place to explain what's the purpose of the new setting (basically, give an example like the one in the original ticket).

zhangtemplar:
@nicolaiarocci updated, thanks!

nicolaiarocci:
rebased, amended, and merged. thanks.

PR #1266, opened 2019-05-15
Fix documents being mutated while computing etag with ignore_fields settings

sunbit:
As copy was used instead of deepcopy, if the ignore_fields setting contains a field that's nested inside others, the function mutated the original while discarding fields to compute the etag as asked.


sunbit:
The travis checks are failing due to this issue already reported here:

https://github.com/pyeve/eve/issues/1267

nicolaiarocci:
rebased and merged. thanks.

PR #1270, opened 2019-05-16
try to fix the total number

hcchien:
For now, the total number is wrong in the _meta. Just try to make it be correct.


hcchien:
the code is removed.

PR #1276, opened 2019-05-29
Fix HEADER_TOTAL_COUNT on emtpy resources

### LINKED ISSUE #1279: Fix HEADER_TOTAL_COUNT on emtpy resources
The condition that avoids returning X-Total-Count when counting is disabled also filters out the case where the resource is empty and count is 0. 

I've opened a PR: #1276

arnauorriols:
the count returned by `app.find` might be None if the count is disabled.
Later in get_internal, the HEADER_TOTAL_COUNT is prepared if the count
is enabled.

This commit fixes a bug that filtered out the HEADER_TOTAL_COUNT when
the resource is empty and the count is 0 because the check is done with a truthy condition. This truthy
condition has been replaced by an explicit `is not None` condition.

Fixes #1279

### DIFF
--- a/eve/methods/get.py
+++ b/eve/methods/get.py
@@ -281,7 +281,7 @@ def _perform_find(resource, lookup):
 
     response[config.ITEMS] = documents
 
-    if count:
+    if count is not None:
         headers.append((config.HEADER_TOTAL_COUNT, count))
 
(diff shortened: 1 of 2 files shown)


nicolaiarocci:
thanks 👍

Issue #1280, opened 2019-06-03
PUT requests doesn't set default values for fields that have one defined


sunbit:
I've been upgrading an old Eve app to latest version, and we just find what we think is a bug introduced back at 2016 whith the 0.8 version. This is particulary bad for us as we're relyng on a field being set from a default value, as we're using that field for filtering the items of the resource.

---

## Expected Behavior

After doing a PUT requests to replace an item, any field that has a default value should be set, as when doing an initial creation POST.

## Minimal settings.py to reproduce the problem
```python
MONGO_HOST = 'localhost'
MONGO_PORT = 27017
MONGO_DBNAME = 'apitest'
IF_MATCH = False
DOMAIN = {
    "test": {
        "schema": {
            "test_field": {"type": "string"},
            "default_field": {"type": "string", "default": "FOO", "allowed": ["FOO"]}
        },
        "resource_methods": ["POST", "GET"],
        "item_methods": ["GET", "PUT"]
    }
}
```
## Minimal app to reproduce the problem

```python
from eve import Eve
app = Eve()

if __name__ == '__main__':
    app.run()
```
## Steps to reproduce it

```
curl -X POST -d '{"test_field": "value"}'  -H 'Content-Type: application/json' http://localhost:5000/test

curl -X GET http://localhost:5000/test/5cf55157f40227844c9b272c

{
  "_updated": "Mon, 03 Jun 2019 16:56:55 GMT",
  "default_field": "FOO",
  "test_field": "value",
 ...
  },
  "_created": "Mon, 03 Jun 2019 16:56:55 GMT",
  "_id": "5cf55157f40227844c9b272c"
}

curl -X PUT -d '{"test_field": "value2"}'  -H 'Content-Type: application/json' http://localhost:5000/test/5cf55157f40227844c9b272c

```

### Actual Behavior

The default value is not set and so effetively deleted from the document after the PUT:

```
curl -X GET http://localhost:5000/test/5cf55157f40227844c9b272c

{
  "_updated": "Mon, 03 Jun 2019 16:58:00 GMT",
  "_created": "Mon, 03 Jun 2019 16:56:55 GMT",
  "test_field": "value2",
  "_id": "5cf55157f40227844c9b272c",
  ...
  }
}
```

I actually identified where it was introduced. The point were the old behaviour was removed from eve.io.mongo.validation and the new one added in eve.validation:

[...]

PR #1282, opened 2019-06-04
Fixes behaviour of PUT requests not setting default values of fields.

sunbit:
* Refactor of "contacts" test data so it honours the fields that should have defaults when created
* Added a new fields in "contacts" schema to be able to properly refactor tests related to default values
* Added tests to cover several scenarios where default values are involved


sunbit:
Fixes #1280.

The most visible effect of that bug was that an existing document that had a default value set on a field after a successful creation, if you tried to replace the document without explicitly sendint that field, the field would disappear from the document.

nicolaiarocci:
Stella work here, thank you @sunbit

sunbit:
You're welcome :), happy to help.

PR #1284, opened 2019-06-06
Pass combined_args to pre_event hooks

### LINKED ISSUE #1283: Lookup argument does not get passed to pre event hook with certain resource urls
### Expected Behavior
If a resource has a url and item_lookup that matches the following:

```python
"url": "user/<regex('[a-f0-9]{24}'):_id>/tasks",
"item_lookup": False,
```

The pre_event hook should provide lookup as the following

```python
def pre_GET_tasks(request, lookup):
    print(lookup)

{'_id': '5ce302e4ef986650a11340fd'}
```

### Actual Behavior

Instead lookup is empty.

I noticed that in the pre_event decorator this issue is solved by combining args with kwargs

```python
combined_args = kwargs
if len(args) > 1:
    combined_args.update(args[1].items())
```

Is there a reason why this bit of code is not moved to the top of the function and the combined_args passed to the pre_event hooks instead of kwargs?

### Environment

* Python version: 3.6
* Eve version: 0.9.1

git-albertomarin:
Fixes #1283 

Where the lookup argument passed to pre_event hooks can have missing parameters for certain user defined urls.

### DIFF
--- a/eve/methods/common.py
+++ b/eve/methods/common.py
@@ -1332,9 +1332,14 @@ def pre_event(f):
         resource = args[0] if args else None
         gh_params = ()
         rh_params = ()
+        combined_args = kwargs
+
+        if len(args) > 1:
+            combined_args.update(args[1].items())
+
         if method in ("GET", "PATCH", "DELETE", "PUT"):
-            gh_params = (resource, request, kwargs)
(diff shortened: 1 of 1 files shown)


nicolaiarocci:
rebased and merged. thanks!

PR #1292, opened 2019-06-30
Add "unique_within_resource" validation rule

### LINKED ISSUE #1291: Shouldn't the scope of the unique validation be the resource, instead of the collection?
After [this change](https://github.com/pyeve/eve/commit/87742343fd0362354b9f75c749651f92d6e4a9c8), the unique attributes must be unique for the entire MongoDB collection. Previously, they had to be unique only for the API resource. This becomes a problem when the datasource is configured with a filter and you have multiple resources sharing the same database collection.

At first we thought this was just a bug. However, in the mentioned commit a comment is introduced indicating that the change is in fact intentional:

> we perform the check on the native mongo driver (and not on
> app.data.find_one()) because in this case we don't want the usual
> (for eve) query injection to interfere with this validation. We
> are still operating within eve's mongo namespace anyway.

However, I fail to understand its motivation. Can you elaborate?

~We don't use the soft-delete feature ourselves, but looking at the code, seems like it should also be affected (ie soft-deleted documents will be taken into account when validating the uniqueness of an attribute).~

arnauorriols:
Add new validation rule `unique_within_resource`.

This new validation rule enforces the uniqueness of an attribute only at API resource level, contrasting with the `unique` rule that enforces uniqueness at database collection level.

Fixes #1291

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1293,6 +1293,17 @@ defining the field validation rules. Allowed validation rules are:
                                 If URRA is not active on the endpoint, this
                                 rule behaves like ``unique``
 
+``unique_within_resource``      The value of the field must be unique within
+                                the resource.
+
+                                This differs from the ``unique`` rule in that
+                                it will use the datasource filter when searching
+                                for documents with the same value for the field.
+                                Use this when the resource shares the database
+                                collection with other resources but their documents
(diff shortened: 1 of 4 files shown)


nicolaiarocci:
thanks.

PR #1324, opened 2019-11-08
Update README funding URL

atwalsh:
Minor update to the README, just noticed while browsing.

Current URL (http://python-eve.org/funding):
image

Updated URL (http://python-eve.org/funding.html):
image


nicolaiarocci:
thanks!

PR #1330, opened 2019-11-26
Some minor style improvements and 2 test fixes

stefaang:
- fixed the 401 or 405 assertion always returning True
- tests were failing on a fresh install without test_user
- I made it a bit more pythonic.

### DIFF
--- a/eve/methods/get.py
+++ b/eve/methods/get.py
@@ -164,12 +164,11 @@ def _perform_aggregation(resource, pipeline, options):
         For example, we have endpoint with a stage like {'$lookup': {'$userId': '$a', '$name': '$b'}}, $a is provided
         but $b is provided as {}. Then the stage will be pruned as {'$lookup': {'$userId': '$a'}}
         """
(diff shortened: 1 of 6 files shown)


nicolaiarocci:
excellent. Tests are in dire need of refactoring (if not a complete rewrite).

nicolaiarocci:
Thanks.

eve/methods/get.py
```diff
@@ -184,7 +183,7 @@ def prune_aggregation_stage(d):
             abort(400, description="Aggregation query could not be parsed.")
 
         for key, value in query.items():
-            if key[0] != "$":
+            if key.startswith("$"):
```
mic-max, line 186: Was this change intended or should it be `if not key.startswith("$"):`

D-stefaang: Hmm not intended.

mic-max: I guess this code path isn't being tested?

nicolaiarocci: Thanks for the review @mic-max. Feel free to PR a fix.

Issue #1332, opened 2019-11-28
Incorrect values deserialization


jordeu:
### Description

Given a resource schema that uses an `anyof` in a `list`, when parsing a value in a field, Eve tries to apply all the possible deserializations but if several deserializations are possible it will end up parsing it using an invalid deserializator.

### Example

Using this resource definition:

```python
from eve import Eve

MONGO_HOST = 'localhost'
MONGO_PORT = 27017
RESOURCE_METHODS = ['GET', 'POST', 'DELETE']
ITEM_METHODS = ['GET', 'PATCH', 'PUT', 'DELETE']

resource = {
    '__description__': __doc__,
    'etag_ignore_fields': ['last_executed'],
    "schema": {
        "test": {
            "type": "list",
            "schema": {
                "anyof": [
                    {
                        "type": "dict",
                        "schema": {
                            "type": {"type": "string", "allowed": ["number"], "required": True},
                            "value": {
                                "type": "number",
                                "required": True
                            }
                        }
                    },
                   {
                        "type": "dict",
                        "schema": {
                            "type": {"type": "string", "allowed": ["boolean"], "required": True},
                            "value": {
                                "type": "boolean",
                                "required": True
                            }
                        }
                    },
                    {
                        "type": "dict",
                        "schema": {
                            "type": {"type": "string", "allowed": ["datetime"], "required": True},
                            "value": {
                                "type": "datetime",
                                "required": True
                            }
                        }
                    },
                    {
                        "type": "dict",
                        "schema": {
                            "type": {"type": "string", "allowed": ["string"], "required": True},
                            "value": {
                                "type": "string",
                                "required": True
                            }
                        }
                    },
                    {
                        "type": "dict",
                        "schema": {
                            "type": {"type": "string", "allowed": ["integer"], "required": True},
                            "value": {
                                "type": "integer",
                                "required": True
                            }
                        }
                    }
                ]
            }
        }
    },
    'url': 'test',
    'resource_methods': ['GET', 'POST'],
    'item_methods': ['GET', 'DELETE', 'PUT', 'PATCH']
}

settings = {'DOMAIN': {'test': resource}}
app = Eve(settings=settings)
app.run()
```

This are valid POST queries that will fail:

```python
# Value is deserialized as an integer
q = {"test": [{ "type": "boolean", "value": True }]}

# Value is deserialized as a boolean
q = { "test": [{ "type": "number", "value": 0 }] }

# Value is deserialized as a datetime
[...]

PR #1333, opened 2019-11-28
Mitigate incorrect values deserialization problem

jordeu:
This pull request mitigates the problem described at #1332.


nicolaiarocci:
howdy! thanks for this. Tests are failing though. Could you look at it?

Issue #1351, opened 2020-01-30
Allow configuration of MONGO_QUERY_WHITELIST


arnauorriols:
Currently the set of allowed operators is fixed in code, which makes sense to avoid unwanted surprises in new versions of MongoDB, but makes the process of including new operators rather rigid. 

For example, I now wanted to experiment with the `$expr` operator, but I don't have any aggregation operator available. The `$eq` operator is missing too. 

I propose to include `eq` in the set of operators allowed (I assume it's just a forgetfulness), and also add a configuration called `MONGO_QUERY_WHITELIST`, where one can add all the extra operators needed. 

What do you think?

I'll push a PR later

nicolaiarocci:
Wow, rather stunned that `$eq` has been missing, and for such a long time.

PR #1352, opened 2020-01-30
Add MONGO_QUERY_WHITELIST global and resource config options

arnauorriols:
Added also $eq as default supported operator. Solves #1351 

I've taken the liberty (without 100% confidence) to add the ..versionchanged tags in the methods changed, and I've used 1.1.0 as version. Was this correct?

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -580,6 +580,15 @@ uppercase.
                                     easily replaced with the (very rich) Mongo
                                     query dialect.
 
+``MONGO_QUERY_WHITELIST``           A list of extra Mongo query operators to allow
+                                    besides the official list of allowed operators.
+                                    Defaults to ``[]``.
+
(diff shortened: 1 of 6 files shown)


nicolaiarocci:
rebased and merged. thanks.

nicolaiarocci:
on the versionchanged tags. I've actually been meaning to remove them all. For starters, they are not accurate (forgot to keep them up to date). Second, they take a lot of vertical space. We have version control for that kind of info anyway.

arnauorriols:
yeah, I know what you mean, it's difficult to maintain, and better nothing than inconsistent. However, I kinda like it when I see it, I think it's useful considering that version control has at most module granularity...

nicolaiarocci:
then, we should make them mandatory on each and every PR (like you did with yours)

PR #1358, opened 2020-02-02
Add coverage report to track what has been tested and what hasn't.

ehiggs:
@nicolaiarocci to take advantage of this you will have to make an account on https://codecov.io/ which links to this github repo.

You will also need to set the CODECOV_TOKEN envvar in your travis build settings. This is a secret token so you need to do it in the web interface instead of adding it to the .travis.yml.

An example report is here: https://codecov.io/gh/ehiggs/eve/branch/add-coverage-report

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,13 @@
+codecov:
+  notify:
+    require_ci_to_pass: no
+    after_n_builds: 1
+coverage:
+  status:
+    project:
+      default:
+        # Require 1% coverage, i.e., always succeed
+        target: 1
+    patch: false
+    changes: false
+comment: off
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[report]
+omit=
+  eve/test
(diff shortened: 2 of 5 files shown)


ehiggs:
Any thoughts?

nicolaiarocci:
Hello. Haven't had the time to look at it yet, I am really sorry. Hope to get back to you within one week or so.

nicolaiarocci:
Hi. I admit I am reluctant to adding more tools to the stack. I recon coverage info would be useful when/if we start refactoring the test suite, something that I have been meaning to do... only to realize that unfortunately I won't have the time for that. Also, to be honest I don't like that codecov does not allow me to sign up with plain old username/pw. Thoughts?

ehiggs:
I personally think all mature projects should report code coverage but seeing as Eve has great coverage already it's not exposing gaping holes in the test cases.

I tried to get it working with coveralls but it was really painful and I found codecov was easier to get setup. Even if you don't want the integration with codecov or coveralls I think having the pytest configuration setup to be able to report coverage is a v. good thing. (obviously I'm biased 😄 )

nicolaiarocci:
Hello, as anticipated, I am incline to not add new tools to the stack, at least for the time being. Thanks.

Issue #1360, opened 2020-02-25
Nested unique constraint doesn't work


elias-garcia:
### Expected Behavior

The `unique` constraint does not work when is inside of a dict or a list, for example. The problem comes from [this line](https://github.com/pyeve/eve/blob/master/eve/io/mongo/validation.py#L95). The resulting query will find only properties in the root document.

The `unique` constraint won't work also when nested inside lists, as the query will try to find by a numeric index instead of the key of the parent schema. 

```json
{
    "dict_test": {
        "type": "dict",
        "schema": {
            "string_attribute": {
                "type": "string",
                "unique": true
            }
        }
    }
}
```

### Actual Behavior

You can create more than one resource of that schema with the same `string_attribute` value.

### Environment

* Python version: 3.7
* Eve version: 1.1

sunbit:
I stepped with this same problem, which made me realize that this wasn't working. The code on #1361 assumes a thing that is the main reason the fix was wrong that is that elf.schema contains the root schema, which doesn't. That code tried to filter the field path validating against the schema definition. My approach on #1433 even if heuristic, i think it's good enough as i can't think of a situation where `self.document_path` containing integers may mean other thing appart from list indexes.

I tried doing a solution checking against self.root_schema, but decided that was not worth the complexity.

PR #1361, opened 2020-02-26
Fix unique constraint in nested attributes

elias-garcia:
This fix resolves the #1360 issue. It adds support for the unique constraint in nested attributes, including lists.


nicolaiarocci:
rebased and merged. thanks.

PR #1362, opened 2020-03-01
Add 403 Forbidden to STANDARD_ERRORS

pedro2555:
This isn't actually an error the Eve Framework will ever raise. As far
as I could tell.

However, separation between authentication and authorization is a must
even for less complex applicatons. 403 is the most adequate response a
server can provide to client whose authentication was accepted, yet for
some reason or another the server will not comply with the request.

The Eve Framework documentation suggests the usage of 403 in, at least
two instances:

 * `abort(403)` in: https://docs.python-eve.org/en/stable/features.html#database-event-hooks
 * The actual documentation for the `STANDARD_ERRORS` config variable,
describes 403 as a supported code, in: https://docs.python-eve.org/en/stable/config.html#global-configuration

### DIFF
--- a/eve/default_settings.py
+++ b/eve/default_settings.py
@@ -131,7 +131,7 @@ VALIDATION_ERROR_AS_LIST = False
 
 # codes for which we want to return a standard response which includes
 # a JSON body with the status, code, and description.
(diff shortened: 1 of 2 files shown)


nicolaiarocci:
Thanks.

Issue #1364, opened 2020-03-04
Unique constraint doesn't work on some nested fields cases


elias-garcia:
### Expected Behavior

Support for the `unique` constraint has some edge cases where it does not work as expected when it's being used in nested schemas. For example when we want a unique `dict` inside a `list`.

```json
{
	"test_list": {
		"type": "list",
		"schema": {
			"type": "dict",
			"unique": true,
			"schema": {
				"first_field": {
					"type": "string"
				},
				"second_field": {
					"type": "string"
				}
			}
		}
	}
}
```

### Actual Behavior

It throws an exception when trying to access an object by a numeric index (the index of each item in the list).

### Environment

* Python version: 3.7
* Eve version: 1.1

PR #1365, opened 2020-03-04
Fix for #1364

elias-garcia:
I've improved the code for the `unique` constraints in nested fields. Now it's working on all kind of schemas. I also think that this implementation is more readable than before for future contributors. I've also implemented some tests and added support to assert to nested validation errors in the tests.


nicolaiarocci:
Hello, thanks for this. There is a failing test on Python 2.7, could you please take a look at it?

PR #1366, opened 2020-03-10
Fix support for dicts without schema rule

arnauorriols:
dicts without schema rule are broken since b8d8fcd5a6d2093ec135a021f5b0f48d2c419dfe

The fix is simple, I've just followed the pattern used for the lists.


nicolaiarocci:
rebased and merged. thanks.

PR #1389, opened 2020-05-07
fix versioning on PATCH when `merge_nested_documents` is disabled

### LINKED ISSUE #1388: disabling merge_nested_documents breaks versioning on PATCH
when saving the version it's using [updated](https://github.com/pyeve/eve/blob/master/eve/methods/patch.py#L230):
```python
insert_versioning_documents(resource, updated)
```
but this updated has the changes from `updates` only if `merge_nested_documents` is enabled on a resource:
```python
if resource_def["merge_nested_documents"]:
    updates = resolve_nested_documents(updates, updated)
    updated.update(updates)  # this one should happen no matter if merging is on or off
```
so if merging is off it saves original again, the first version will be duplicated and the latest will be always missing.

petrjasek:
fix #1388

### DIFF
--- a/eve/methods/patch.py
+++ b/eve/methods/patch.py
@@ -212,7 +212,8 @@ def patch_internal(
 
             if resource_def["merge_nested_documents"]:
                 updates = resolve_nested_documents(updates, updated)
-                updated.update(updates)
+
+            updated.update(updates)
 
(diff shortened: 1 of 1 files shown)


nicolaiarocci:
thanks.

PR #1398, opened 2020-06-04
Add missed condition when projection is disabled per domain

ride90:
I think there is a missed part in the condition for projection.
I've added a projection flag from the resource.


ride90:
Hi @nicolaiarocci , please, review this PR.

nicolaiarocci:
thanks!

Issue #1411, opened 2020-09-18
Race condition in PATCH on newly created documents with clustered mongo


ehiggs:
When using a replicated set of mongo instances, it's normal to have the default client be configured to` PREFER_SECONDARY`. When performing a `patch` (or `patch_internal`), the code will call `get_document`, which then calls `find_one` which can be to the secondary. If the write concern is not total (also common) then the retrieved document state can be stale. This is a problem when one wants to insert data immediately, do some quick processing, and then patch the document to say the processing is done: the patch request might find that 'there is no document'.

### Expected Behavior

PATCH on newly created documents should work.

```python
post_internal(...)
patch_internal(...)
```

### Actual Behavior

'No such document'

### Environment

* Python version: 3.7
* Eve version: 1.1.1

nicolaiarocci:
Fixed with #1412

PR #1412, opened 2020-09-18
Allow mongo_options to be passed to post, patch, and get_document.

ehiggs:
This is so users can control ReadPreference and WriteConcern. This should address #1411


nicolaiarocci:
Excellent. Thank you.

PR #1414, opened 2020-09-22
Fix the mongo_options work by passing the options into find_one

ehiggs:
This is done by passing the options directly into the `eve.Mongo` `find_one` method.
The work doesn't cover adding arguments to find_one_raw or find since
they aren't needed to fix the bug.

I explored adding a fluid type interface for with_optios to the eve
PyMongos/Mongo/DataLayer classes but it was really clunky and I'd be
worried that such a fundamental change wouldn't be appropriate here.

### DIFF
--- a/eve/io/base.py
+++ b/eve/io/base.py
@@ -163,6 +163,7 @@ class DataLayer(object):
         req,
         check_auth_value=True,
         force_auth_field_projection=False,
+        mongo_options=None,
         **lookup
     ):
         """Retrieves a single document/record. Consumed when a request hits an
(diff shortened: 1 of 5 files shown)


nicolaiarocci:
howdy, looks like there are some linting issues. could you please address them before I can merge? thanks.

ehiggs:
Build failure appears to be due to black high had previously failed in an unrelated test.

ehiggs:
The failed test seems to be flaky as it worked on my branch: https://travis-ci.org/github/ehiggs/eve/builds/732945620

@nicolaiarocci could you retry the test?

ehiggs:
Used the `git commit --edit --amend` incantation to nudge Travis.

PR #1419, opened 2020-10-28
Instantiate GridFS object with 'disable_md5=True' (fix #1410)

### LINKED ISSUE #1410: GridFS support of md5 is deprecated
Per the mongodb GridFS documentation at https://docs.mongodb.com/manual/core/gridfs/#files.md5

The md5 field is deprecated and should no longer be used. The pymongo gridfs driver has an argument "disable_md5" which defaults to "False". Eve should default that argument to "True". As the documentation suggests, clients can calculate their own md5 or any other hash and place it in the metadata field within gridfs.

If you don't wish to default this to "True", then please provide a mechanism to set this when defining the domain.

### Expected Behavior

GridFS class is instantiated with `disable_md5=True`

```python
DOMAIN = {
  'myrecords': {
    'schema': {
      'file': {
        'type': 'media'
      }
    }
  }
}
```

### Actual Behavior

GridFS is instantiated with disable_md5=False (default value when argument not provided) which causes warnings in other mongodb tooling about the deprecation of md5 support. 

```pytb
Paste the full traceback if there was an exception.
```

### Environment

* Python version: 3.8.2
* Eve version: 1.1.2

alexmisk:
Hi!

Do you think any tests needed for this fix?

### DIFF
--- a/eve/io/mongo/media.py
+++ b/eve/io/mongo/media.py
@@ -59,7 +59,7 @@ class GridFSMediaStorage(MediaStorage):
 
         px = driver.current_mongo_prefix(resource)
         if px not in self._fs:
-            self._fs[px] = GridFS(driver.pymongo(prefix=px).db)
+            self._fs[px] = GridFS(driver.pymongo(prefix=px).db, disable_md5=True)
         return self._fs[px]
 
     def get(self, _id, resource=None):


nicolaiarocci:
thanks!

PR #1426, opened 2020-11-30
corrected variable name

rahulsalgare:
inside example section of insert events, corrected the variable name in 'before_insert' function


nicolaiarocci:
thanks!

PR #1433, opened 2021-01-21
Fix nested unique fields validation

sunbit:
The previous attempt to fix this was not working with nested unique fields


sunbit:
Sorry for this, i tried to delete it but i can't see how, good PR in #1434

PR #1435, opened 2021-01-22
Fix unique field query construction when field is nested

### LINKED ISSUE #1434: Nested unique fields still don't work
### Expected Behavior

When adding unique fields not only as root fielkds but nested inside an arbitrary number of dicts, lists or combination of both, the unique validation should work. 

See the [test schemas](https://github.com/sunbit/eve/blob/fix_unique_query/eve/tests/test_settings.py#L305-L338) that when used the expected behaviour is not honored.

### Actual Behavior

Exception caused by KeyErrors in the actual code that tries to fix this in 1.1.1

### Environment

* Python version: any
* Eve version: 1.1.1+

This issue was also reported on #1361 but that merged fix doesn't work because:
- code assumes that self.schema contains the root schema, which is not that way
- code ignores de nested `schema` sections of a schema, as if the schemas is equal in structure as the document.
 I uc

sunbit:
Fixes #1434 and #1360

### DIFF
--- a/eve/io/mongo/validation.py
+++ b/eve/io/mongo/validation.py
@@ -94,19 +94,27 @@ class Validator(Validator):
         .. versionadded:: 0.6
         """
         if unique:
-            schema = self.schema
-            attribute_path = self.document_path + (field,)
-            temp_path = [attribute_path[0]]
 
-            for i, path in enumerate(attribute_path[:-1]):
-                schema = schema[path]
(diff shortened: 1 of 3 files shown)


sunbit:
Not sure why the travis job failed, seems something completely unrelated to the changes, any idea?

nicolaiarocci:
thanks @sunbit

PR #1437, opened 2021-01-28
Add Python 3.9 support

eumiro:
Hi @nicolaiarocci, would you consider dropping support for Python 2.7 and 3.5 (after EOL)?


nicolaiarocci:
Thanks @eumiro.

I am OK with dropping 3.5 right away. I would drop 2.7 but long term, maybe when we hit 2.0.

PR #1438, opened 2021-02-06
Drop Python 3.5 support and update pypy3 version

eumiro:
As discussed in #1437, drop Python 3.5 support (and looking forward to dropping 2.7).

And update pypy3 to the latest version in travis.

How about adding GitHub Actions CI?


nicolaiarocci:
Moving away from Travis would be ideal.

PR #1439, opened 2021-02-07
Add GitHub Actions CI

eumiro:
First let's see whether it runs …

PR #1444, opened 2021-02-23
Drop Travis CI

eumiro:
As mentioned in #1439, add GitHub Actions CI badge to the docs and drop Travis CI entirely.


nicolaiarocci:
@eumiro would mind taking a look at why tests are currently failing on CI? See [requirement](https://github.com/pyeve/eve/actions)

PR #1445, opened 2021-03-01
Prepare for Python3

eumiro:
Two independent commits:
- `from __future__ import division` and then apply it to two integers where `a / b` returns a float
- simplify multiple `isinstance` calls


nicolaiarocci:
thanks!

nicolaiarocci:
@eumiro with the advent of pymongo 4 (also see #1461) which only supports Python 3, it might be the right time for the switch. Would you be interested in tackling the issue?

PR #1449, opened 2021-04-01
Fix typo in config

itsjef:
`VALIDATION_ERROR_AS_STRING` doesn't exist in the source code so set it to `True` won't have any effect.

It must be `VALIDATION_ERROR_AS_LIST` instead.


nicolaiarocci:
aww sorry for being so late

PR #1463, opened 2022-01-22
Update docs and tests regarding pagination of empty resources

kinuax:
According to the code, the presence of the pagination data (the `_meta` field) in the payload is only determined by the pagination setting, no matter the number of documents returned. Docs and tests are updated in order to clarify it.

Missing Black code-style changes are applied so that `pre-commit` passes.


nicolaiarocci:
rebased and merged. Thanks.

PR #1471, opened 2022-03-11
Remove warnings from Cerberus

Kludex:


nicolaiarocci:
We should keep the pin on cerberus 1.x stable.

nicolaiarocci:
Yeah, apparently the new `ObjectId` custom type is returning `false` on valid objectid string values. I suspect it has to do with the values being strings. Same error actually comes with `Poligon`, `Feature`, and other, more comples geo-types.

Kludex:
I've found the issue.

On the previous implementation, we had a `type_handler` (line 1563). Now our validation goes through the other path, as we have a `type_definition` (line 1557). And well... The expression really doesn't return `True`, so `match` will be `False` on line 1559. - The snippet below is from Cerberus 1.3.4. 

image

How should I proceed?

nicolaiarocci:
Yeah, basically what happens is this:

```
    >>> isinstance('5349b4ddd2781d08c09890f3', ObjectId)
    false
```

Which of course makes sense from Cerberus' `type` rule perspective. As I see it we're left with three options:

1) Abandon some (most) of current Eve default types, replacing them with [`check_with`](https://docs.python-cerberus.org/en/stable/validation-rules.html#check-with)-ready rules. This would be a major breaking change.

2) Accept deprecation warnings, at least for them time being. People who get their CI pipelines in trouble with them can always [opt-out](https://docs.python-cerberus.org/en/stable/validation-rules.html#check-with) after all.

3) Wait for Cerberus to eventually 'power-up' the current `type` rule, so that it can also accepts custom handlers. I'm summoning @funkyfuture on this one. I'm not sure I'd agree with or like such a move however (a string representing an objectid value is not an obectid, after all.) 

Option 1 would better position Eve for an eventual, future major release of Cerberus (v2). Option 2 is of course the easy way out. All things considered, I'd pick option 2. Thoughts?

Kludex:
Just to be sure: option 2 implies on not proceeding with a fix for the issue this PR tries to solve, correct?

In any case, I'm fine with any solution. I don't have strong feelings, as I've started to use eve recently. I'm willing to work on anything required to have eve released.

funkyfuture:
can someone point me to the code where these Python types are defined? @Kludex, have you looked into the usage of the third parameter of `TypeDefinition`?

regarding, a Cerberus 2, leaving out the sad part of the story, no matter when it will be available, it would certainly require a new major version of Eve too. so, proposal 3 is similar to 1.

Kludex:
> can someone point me to the code where these Python types are defined?

Hmm... The ones on the CI or...?

> have you looked into the usage of the third parameter of TypeDefinition?

How would the third parameter help here?

nicolaiarocci:
> Just to be sure: option 2 implies on not proceeding with a fix for the issue this PR tries to solve, correct?

Correct.

funkyfuture:
> Hmm... The ones on the CI or...?

i meant the classes of which i found most in `eve.io.geo`.

> How would the third parameter help here?

i don't know, i wanted to check whether you had already considered that as feasible for a solution. that parameter allows to define Python super- and sub-classes that shall not be considered as the defined type. e.g. [here](https://github.com/pyeve/cerberus/blob/1.3.x/cerberus/validator.py#L138) a boolean is not to be considered as number.

i see a general bad pattern in these snippets:

```python
    def _validate_type_point(self, value):
        try:
            Point(value)
            return True
        except TypeError:
            pass
```

here, that value isn't tested for its type, but *whether it can* be coerced into another type. it would be correct to write a schema like this though:

```python
{
    "field": {
        "coerce": Point,
        "type": Point,
    }
}
```

but maybe y'all wanna [just filter](https://docs.python.org/3/library/warnings.html#warning-filter) what jangles you?

Kludex:
> but maybe y'all wanna [just filter](https://docs.python.org/3/library/warnings.html#warning-filter) what jangles you?

I can do that. If the idea is to bump a major on eve later on, then it doesn't look a bad idea, I guess?

nicolaiarocci:
Let's take the current `objectid` type as an example. I would look into droping `_validate_type_objectid()` in the mongo validator class, then I'd add a replacement function that can be used with Cerberus' `check_with`. So, [this example](https://docs.python-eve.org/en/stable/validation.html#custom-data-types):

```
'schema': {
    'owner': {
        'type': 'objectid',
        'required': True,
    },
}
```
Would need to be updated to:
```
'schema': {
    'owner': {
        'check_with': is_valid_objectid,
        'required': True,
    },
}
```
Builtin check_with function are not strictly required, but would be nice to have. Can't look at it right now, but I'd wild guess that the `validate_filters` function you linked will work out of the box (if not, we'll then look into it.)

- the same would happen to other Cerberus-deprecated types (geo, maybe others?) 
- docs would need careful revision/edit
- new bultin functions would need test coverage
- existing tests will also need updates

Sounds reasonable?

funkyfuture:
> Sounds reasonable?

from a distance, yes.

where possible i'd recommend `regex` (and a string type check) over `check_with` for better performance.

be reminded that a [rules set](https://docs.python-cerberus.org/en/stable/schemas.html#registries) may be an option to avoid duplication.

just ping me when i shall have a look at a patch proposal.

nicolaiarocci:
Yes, kind of. Maybe move the built-in function out of the Validator class. 

However, I took some time to dig into it. I experimented a little bit with Geo classes and types, and then with ObjectIds. It's not a trivial task. Eve is so rooted into old-style Cerberus type system that the work required is not worth it. Cerberus 1.x is serving Eve just fine, and breaking all Eve deployments just  to achieve the removal of some warnings that can be otherwise silenced/filterd (see above links) would be unjustified.

Thanks @Kludex and @funkyfuture for lending a hand. It was worth investigating this. I'm closing this draft PR and the linked ticket, at least for the time being.

PR #1473, opened 2022-04-09
render.py: use markupsafe since werkzeug has removed escape from its utils

dasbooss:
https://github.com/pallets/werkzeug/commit/22d1e9ac13829b83347107a9b4d77072a8e1af6a

Fixes this error:
AttributeError: module 'werkzeug.utils' has no attribute 'escape'

### DIFF
--- a/eve/render.py
+++ b/eve/render.py
@@ -15,6 +15,7 @@ import time
 import datetime
 import simplejson as json
 from werkzeug import utils
+from markupsafe import escape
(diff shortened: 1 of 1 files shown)


heharkon:
Ah, should have tried with the dev version... Thanks!

nicolaiarocci:
rebased and merged. the CI crash was not due to this specific change, see #1475. Thanks!

PR #1482, opened 2022-07-21
fix auth using embedded username/password in uri

smeng9:
Closes https://github.com/pyeve/eve/issues/1478

### DIFF
--- a/eve/io/mongo/flask_pymongo.py
+++ b/eve/io/mongo/flask_pymongo.py
@@ -64,16 +64,29 @@ class PyMongo(object):
             host = app.config[key("URI")]
             # raises an exception if uri is invalid
             mongo_settings = uri_parser.parse_uri(host)
+
+            # extract username and password from uri
+            if mongo_settings.get("username"):
+                client_kwargs["username"] = mongo_settings["username"]
+                client_kwargs["password"] = mongo_settings["password"]
+
+            # extract default database from uri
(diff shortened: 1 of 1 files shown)


nicolaiarocci:
Hi, thanks for this. 

At quick glance, I'd say that If `USERNAME` is set in config, it will override the URI values (see [here](https://github.com/pyeve/eve/blob/4eaf0a34721c20c7cf20d689088c3dc72d223ac1/eve/io/mongo/flask_pymongo.py#L102)). I'm thinking it should be the other way around. Thoughts?

smeng9:
Hi @nicolaiarocci , I would think there should only be one set username/password, either in uri, or as config username,password. So if user uses the username/password in uri, they should not manually set username/password somewhere else.

gwainer:
Hello, I'm looking forward to use eve 2.0 but this issue is holding us. What about adding [only this block](https://github.com/pyeve/eve/pull/1482/commits/4eaf0a34721c20c7cf20d689088c3dc72d223ac1#diff-9e2bc5c13e251bf695cc491a8c4a8352f5473020cec4ddb076d15cea46f4092bR79-R81) in the PR and move on? The username part I don't see as related to the issue here. I can help if needed.

nicolaiarocci:
rebased and merged. thanks.

nicolaiarocci:
PS: the CI crash was due to Flask 2.2 breakage

Issue #1486, opened 2022-09-21
etag generation fails if 'uuidRepresentation' is not set in MONGO_OPTIONS


tgm-git:
**This issue tracker is a tool to address bugs in Eve itself.
Please use Stack Overflow for general questions about using Eve or issues not
related to Eve (see http://python-eve.org/support).**

If you'd like to report a bug in Eve, fill out the template below. Provide
any any extra information that may be useful / related to your problem.
Ideally, create an [MCVE](http://stackoverflow.com/help/mcve), which helps us
understand the problem and helps check that it is not caused by something in
your code.

---

### Expected Behavior

When 'uuidRepresentation' is not set in MONGO_OPTIONS generating etags should still be possible.

```python

-- app.py
my_settings = {
    'MONGO_HOST': 'localhost',
    'MONGO_PORT': 27017,
    'MONGO_DBNAME': 'the_db_name',
    'DOMAIN': {'contacts': {}}
}

app = Eve(settings=my_settings)
app.run()


-- some_test.py
from eve import post_internal
post_internal('contacts', {'a_valid_contact': 'this must be'}, skip_validation=True)
```

### Actual Behavior

Since no 'uuidRepresentation' is present in MONGO_OPTIONS, the default option is selected, which is an int, not a string. A simple fix would be chaning line 350 in 'eve/utils.py' from `config.MONGO_OPTIONS.get("uuidRepresentation", UuidRepresentation.STANDARD)` to `config.MONGO_OPTIONS.get("uuidRepresentation", 'standard')`

or add a field 4 to the uuid_map on line 343: 

```
uuid_map = {
            "standard": UuidRepresentation.STANDARD,
            4: UuidRepresentation.STANDARD,
            "unspecified": UuidRepresentation.UNSPECIFIED,
            "pythonLegacy": UuidRepresentation.PYTHON_LEGACY,
            "csharpLegacy": UuidRepresentation.CSHARP_LEGACY,
            "javaLegacy": UuidRepresentation.JAVA_LEGACY,
        }
```

```pytb
    def uuid_representation_as_string():
        uuid_map = {
            "standard": UuidRepresentation.STANDARD,
            "unspecified": UuidRepresentation.UNSPECIFIED,
            "pythonLegacy": UuidRepresentation.PYTHON_LEGACY,
            "csharpLegacy": UuidRepresentation.CSHARP_LEGACY,
            "javaLegacy": UuidRepresentation.JAVA_LEGACY,
        }
[...]

PR #1487, opened 2022-09-21
Fix issue  #1486

tgm-git:
This solution seems to fit the current code best, but alternatively one could get the value earlier and return  `UuidRepresentation.STANDARD` early if the value is of a non-string type.

This PR is a fix to [issue #1486](https://github.com/pyeve/eve/issues/1486)


tgm-git:
I am not sure if I should be comparing with the master branch, but this fix is essentially a hotfix.

nicolaiarocci:
thanks!

tgm-git:
No worries :) your project is saving me hours of headaches

PR #1488, opened 2022-09-22
[wip] Invalid bearer token should be 401 instead of 500

thepieterdc:
When passing an invalid authorization token header, an internal server error is raised due to the `split()` function assuming that the header will always contain a space -- whereas this should be a client error instead.

Contains tests


thepieterdc:
Seems to be fixed already in 2.0.0 :)

PR #1493, opened 2022-11-03
python 3 updates

marksmayo:
and some simplifications of conditional statements.

### DIFF
--- a/eve/io/mongo/__init__.py
+++ b/eve/io/mongo/__init__.py
@@ -10,7 +10,7 @@
     :license: BSD, see LICENSE for more details.
 """
 
+from eve.io.mongo.media import GridFSMediaStorage
 # flake8: noqa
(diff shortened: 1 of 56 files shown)


eve/io/mongo/__init__.py
```diff
@@ -10,7 +10,7 @@
     :license: BSD, see LICENSE for more details.
 """
 
+from eve.io.mongo.media import GridFSMediaStorage
```
nicolaiarocci, line 13: tests are failing because of this line. If you move it back to the bottom, they'll pass (imports order is important in this case.)

nicolaiarocci (changes requested):

nicolaiarocci:
alright i went ahead and applied the fix myself. Thanks for your contribution!

marksmayo:
@nicolaiarocci sorry for the delayed response. Thanks for sorting that!

PR #1507, opened 2023-07-07
Moving tests folder to root

guillaumeLepape:
Mainly to avoid including it in python wheel


guillaumeLepape:
I don't know which command you are using to release your build. I tried with `make wheel` and install it in a python env. Seems ok.
image

nicolaiarocci:
thanks!

PR #1547, opened 2025-02-26
Correct on_deleted_resource call signature in docs

sbe-visma:
I was confused from the docs about what the "item" in the on_deleted_resource would be, but after looking at [the code](https://github.com/pyeve/eve/blob/172f3f251bc34d7c59cd72c9b67873b99ba0f486/eve/methods/delete.py#L254-L255) it seems this is an oversight in the docs, those functions are only called with the resource name (or nothing if the hook is resource specific). This updates the docs to match


nicolaiarocci:
rebased and merged, thanks.

PR #1548, opened 2025-02-26
Update README.rst

aziel18:
update


nicolaiarocci:
Not interested, thank you. Besides, this looks like it was geneated with some kind of LLM.

Issue #1552, opened 2025-05-17
`field_definition` impacts negatively the performance of `build_response_document`


aeurielesn:
In the course of profiling one of our endpoints, `build_response_document` seemed to be taking a surprising amount of time.

```sh
PATH: '/endpoint'
         9168960 function calls (9168445 primitive calls) in 2.798 seconds

   Ordered by: internal time, call count

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     4897    0.021    0.000    2.112    0.000 eve/methods/common.py:603(build_response_document)
   362378    0.059    0.000    0.784    0.000 eve/methods/common.py:702(field_definition)
     4897    0.074    0.000    1.052    0.000 eve/methods/common.py:743(resolve_data_relation_links)
```
(I took the liberty of summarizing the profiling output down to the functions of interest here)

Considering `field_definition` is strictly dependent of `config.DOMAIN` which it is static at runtime, caching `field_definition` seems to greatly alleviate this problem.

```diff
diff --git a/eve/methods/common.py b/eve/methods/common.py
index f8d46b68..0bd2f4cd 100644
--- a/eve/methods/common.py
+++ b/eve/methods/common.py
@@ -15,7 +15,7 @@ import time
 from collections import Counter
 from copy import copy
 from datetime import datetime, timezone
-from functools import wraps
+from functools import cache, wraps
 
 import simplejson as json
 from bson.dbref import DBRef
@@ -699,6 +699,7 @@ def resolve_resource_projection(document, resource):
         del document[field]
 
 
+@cache
 def field_definition(resource, chained_fields):
     """Resolves query string to resource with dot notation like
     'people.address.city' and returns corresponding field definition
```

As shown by a newest summary of the profiling output:

```sh
PATH: '/endpoint'
         5504982 function calls (5504602 primitive calls) in 1.547 seconds

   Ordered by: internal time, call count

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     4897    0.018    0.000    0.565    0.000 eve/methods/common.py:603(build_response_document)
   362378    0.042    0.000    0.042    0.000 eve/methods/common.py:702(field_definition)
     4897    0.066    0.000    0.310    0.000 eve/methods/common.py:743(resolve_data_relation_links)
```

nicolaiarocci:
nice find, would you be willing to submit a pull request? I'd be happy to review it.

aeurielesn:
[...]

PR #1553, opened 2025-05-18
Caches `field_definition` calls to improve `build_response_document` performance

aeurielesn:
- Issue: #1552.


nicolaiarocci:
Awesome, thank you.

PR #1556, opened 2025-06-15
Fixes #1551

### LINKED ISSUE #1551: ensure_mongo_indexes issue with mongo 8
There is new behaviour when creating indexes in mongo 8 that causes a bad interaction with `ensure_mongo_indexes`. If an index is hidden, then the create index returns an `IndexOptionsConflict` error causing Eve to mistakenly think it needs to recreate the index.

The following code works in mongo 7 but fails in mongo 8:
```
db["mycollection"].createIndex({field1:1, field2:1}, {name: "myindex"})
myindex
db["mycollection"].hideIndex("myindex")
{ hidden_old: false, hidden_new: true, ok: 1 }
db["mycollection"].createIndex({field1:1, field2:1}, {name: "myindex"})
MongoServerError[IndexOptionsConflict]: An equivalent index already exists with the same name but different options. Requested index: { v: 2, key: { field1: 1, field2: 1 }, nam...
```

This means that the following code fails in _create_index:

```
   for coll in colls:
        try:
            coll.create_index(list_of_keys, **kw)
        except pymongo.errors.OperationFailure as e: # <------ this is hit because we get an IndexOptionsConflict
            if e.code in (85, 86):
                # raised when the definition of the index has been changed.
                # (https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err#L87)

                # by default, drop the old index with old configuration and
                # create the index again with the new configuration.
                coll.drop_index(name)
                coll.create_index(list_of_keys, **kw)
            else:
                raise
```

Workaround(s):
1. Don't hide indexes if one is using the mongo_index functionality of eve.
2. If something needs to be hidden, comment the index out of the eve mongo_indexes section of the domain schema.

y3h7rv:
Closes #1551.

### Summary

This PR addresses an issue where Eve would unnecessarily drop and recreate an index when encountering a benign `IndexOptionsConflict` error (code 85) on MongoDB 8.0+. This behavior, while not crashing the application, is inefficient and potentially risky on large collections.

### The Problem (Old Behavior)

In MongoDB 8.0+, attempting to `create_index` on an index that already exists but is `hidden` now correctly raises an `IndexOptionsConflict` error. The previous logic in `_create_index` caught this error but treated it identically to a genuine index definition conflict (e.g., different keys or options). This led to a destructive `drop_index` followed by a `create_index` operation.

### The Solution (New Behavior)

This patch refactors the exception handling block within `_create_index` to be more intelligent. It now proactively checks the reason for the `IndexOptionsConflict`:

* It uses `coll.index_information()` to get the details of the existing index.
* If it confirms that the index definition (keys, options, etc.) matches and the only conflict is the `hidden` state, it now simply un-hides the index using a `collMod` command instead of recreating it. This is a much safer and more efficient approach.
* For all other index conflicts (e.g., mismatched keys or a change in the `unique` option), the original drop-and-recreate behavior is correctly preserved.

### Testing

* A new regression test has been added to specifically simulate this hidden index scenario, ensuring the new logic is correctly triggered.
* The new test case fails without this patch and passes with it.
* All existing tests continue to pass locally via `tox -e linting,py310,py39`.

### DIFF
--- a/eve/io/mongo/mongo.py
+++ b/eve/io/mongo/mongo.py
@@ -27,8 +27,13 @@ from werkzeug.exceptions import HTTPException
 from eve.auth import resource_auth
 from eve.io.base import BaseJSONEncoder, ConnectionException, DataLayer
 from eve.io.mongo.parser import ParseError, parse
-from eve.utils import (config, debug_error_message, str_to_date, str_type,
-                       validate_filters)
+from eve.utils import (
+    config,
+    debug_error_message,
+    str_to_date,
+    str_type,
+    validate_filters,
+)
 
 from ...versioning import versioned_id_field
 from .flask_pymongo import PyMongo
@@ -293,7 +298,7 @@ class Mongo(DataLayer):
         check_auth_value=True,
         force_auth_field_projection=False,
         mongo_options=None,
-        **lookup
+        **lookup,
     ):
         """Retrieves a single document.
 
@@ -1137,6 +1142,10 @@ def _create_index(app, resource, name, list_of_keys, index_options):
     For example:
         {"sparse": True}
 
+    .. versionchanged:: 0.8.2
+       Add intelligent handling for hidden indexes and improved error handling
+       for index conflicts. Check existing index properties before drop/recreate.
+
     .. versionchanged:: 0.8.1
        Add support for IndexKeySpecsConflict error. See #1180.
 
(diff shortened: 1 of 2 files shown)


nicolaiarocci:
I think the workarounds suggested by @ehiggs are perfectly fine. As I understand them, [hidden indexes](https://www.mongodb.com/docs/manual/core/index-hidden/) use case is that of a temporary experiment anyways.

